PR #1, opened 2024-01-02
Make the use of a full PKI tree configurable for i.MX8

rborn-tx:
This adds the handling for the "CA flag" to i.MX8 devices (for both HABv4 and AHAB); more detail is provided in the individual commit messages.

Notice that the first two (small) commits are actually unrelated to the "CA flag"; the former is a small fix for a bug on a recently merged commit and the latter was needed due to a BSP change on recipe `imx-mkimage` that broke the parsing of the `mkimage` log.


sergioprado:
Looks good to me!

PR #3, opened 2024-01-17
Make room for compiling the SPL with HAB support

rborn-tx:
This basically adds two new commits that aim to reduce the size of the SPL binary so that HAB support can be compiled in (which is required to the secure boot images). With this, the build failure on the verdin-imx8mm should be solved for the moment.

~**NOTE**: I'm leaving this as draft for now until some discussion with the BSP team is finished.~


sergioprado:
Does it make sense to disable these config options for all machines? I understand only mx8m is affected for now, but since those features are not really required, we might want to disable it for all machines, and prevent other machines to have the same problem in the future.

In that case, I would create another U-Boot configuration fragment only to disable features not required in U-Boot/SPL.

jsrc27:
I understand the objective was to disable unused configs therefore reducing the bloat/size. How exactly did you decide on these configs as the ones to disable? Did you just look at which configs had the biggest impact on size or was there some other logic to your choice?

rborn-tx:
@sergioprado

> Does it make sense to disable these config options for all machines? I understand only mx8m is affected for now, but since those features are not really required, we might want to disable it for all machines, and prevent other machines to have the same problem in the future.

They way I see it, the changes introduced with this PR are temporary just to solve the current build issue; if they weren't then I'd definitely answer your question above with an "yes". But let me give some reasons I believe this solution should be temporary:

1. The build issue was introduced due to a vulnerability fix added to the `meta-toradex-nxp` layer. Before that fix the disabling of specific hashing algorithms (SHA1, SHA384, ...) -- as being added in the present PR -- wasn't needed at atll since the algorithms wouldn't be linked in the first place. Thus, the fix in the present PR is dependent on changes done in another layer.
2. The disabling for the "USB gadget support" was **chosen arbitrarily** simply based on the code footprint it adds. By doing this we might be ignoring requirements for the BSP layer (for example, maybe that feature was enabled due to some customer request) rather than inheriting them as we should (right?). Besides, this disabling is only effective because we know the feature is currently enabled in another layer and again the solution here depends on another layer.

So, to avoid both issues above and in general to avoid implicit dependencies between layers, I believe the ideal solution in terms of maintenance would be for each layer to only enable features it needs and never remove features since *normally* the absence of a feature shouldn't be a requirement for something to work.

Of course, if we didn't have control of the BSP layers then we wouldn't have better options but to make these arbitrary removals. However, since we can ask for changes to the BSP layers, I thought the solution was to ask the BSP team to ensure HAB can be enabled in upper layers (i.e. they would make sure there is room for adding that feature). So, once that is implemented (for example, they could do it be disabling other features they deem less important) we could roll back the changes added with the present PR.

Does all of that make sense?

P.S.: @jsrc27 Item (2) above should also answer your question, but I can add that I tried to find a single unused feature that added lots of code and that seemed to be the "USB gadget support".

sergioprado:
@rborn-tx thanks, that makes sense indeed.

Any particular reason to split this in two commits instead of one? This is because you think we might revert just one of them later?

Also, since this is provisory, I would recommend opening an issue here so we don't forget later that we need to revert this.

rborn-tx:
@sergioprado 

> Any particular reason to split this in two commits instead of one? This is because you think we might revert just one of them later?

Yes. One commit disables an arbitrary config option while the other disables config options related to the vulnerability fix. So, depending on how things are solved at the BSP side we might want to remove one or the other or ideally both.

> Also, since this is provisory, I would recommend opening an issue here so we don't forget later that we need to revert this.

Good idea. I'll do that. Thanks!

sergioprado:
> Good idea. I'll do that. Thanks!

Great, thanks!

Let's see if @jsrc27 has any other comments before we merge this.

jsrc27:
Given the explanation, all looks good from my side then. Since the changes here are temporary I guess the choice of configs does not matter all that much in the long-term.

PR #5, opened 2024-01-30
Add support for signing FIT images on AM62

sergioprado:
This PR adapts meta-toradex-security so it can be used on SoCs other than iMX and implements FIT image siging for AM62 based hardware.

Tested on Verdin AM62 and Verdin iMX8MP.


recipes-bsp/u-boot/u-boot-fit-signature-ti.inc
```diff
@@ -0,0 +1,9 @@
+do_deploy:prepend() {
+    # when the FIT image is generated and signed, the public key is included
+    # in u-boot.dtb, and the U-boot image needs to be regenerated so the new
+    # DTB is used
+    eval $(cat include/config/auto.conf | grep ^CONFIG_DEFAULT_DEVICE_TREE)
+    install -m 0644 u-boot.dtb arch/arm/dts/${CONFIG_DEFAULT_DEVICE_TREE}.dtb
+    oe_runmake ${UBOOT_MAKE_TARGET}
+    install -m 0644 u-boot.img u-boot-${UBOOT_CONFIG}.img
```
rborn-tx, line 8: @sergioprado A couple questions:

* Shouldn't we refer to the generated binary via `${UBOOT_BINARY}`?

```
    install -m 0644 u-boot.img u-boot-${UBOOT_CONFIG}.img
                    ^^^^^^^^^^
```

* Could you clarify the purpose of making this copy of `u-boot-img` into the build directory? Shouldn't the generated file go to a deployment directory?

rborn-tx (commented):
Finished a pass; I've raised some questions but nothing major.

jsrc27 (approved):
I don't have anything additional to add. Looks fine to me for the most part.

recipes-bsp/u-boot/u-boot-fit-signature-ti.inc
sergioprado: > Shouldn't we refer to the generated binary via ${UBOOT_BINARY}?
Makes sense! PR updated.

> Could you clarify the purpose of making this copy of u-boot-img into the build directory? Shouldn't the generated file go to a deployment directory?
Yeah, this is just messy. The amount of different U-Boot images generated is really confusing. The thing is that the compile task will generate both `u-boot.img` and `u-boot-sd.img` files (`UBOOT_CONFIG=sd`). Then the default deploy task will deploy both images to `DEPLOYDIR`. In the end, `u-boot-sd.img` is used in the Easy Installer image. So in this task, we are just rebuilding the U-Boot image to get the new DTB with the public key and updating `u-boot-sd.img`. Since the default deploy task will deploy those files, we don't need to deploy them, just rebuilding is enough.

PR #9, opened 2024-02-21
ti-k3-secdev: apply append only if secure boot is enabled

sergioprado:
The changes to update the signing keys should only be applied if secure boot is enabled.


rborn-tx (approved):
LGTM

jsrc27 (approved):
LGTM as well.

PR #11, opened 2024-02-28
classes: introduce tdxref-signed class

sergioprado:
This class is responsible for creating a verified BSP reference image with full chain-of-trust support.

It inherits `tdx-signed` to sign boot artifacts (bootloader and FIT image) and `tdx-signed-dmverity` to boot a signed rootfs image with dm-verity.

The `tdx-signed-dmverity.class` implementation is currently incomplete, and for now it just builds and boots the rootfs via a ramdisk image (required for the dm-verity signature verification).

Build tested on Verdin AM62, Verdin iMX8MP and Colibri iMX6.

Runtime tested on Verdin iMX8MP and Colibri iMX6.

### DIFF
--- a/classes/tdx-signed-dmverity.bbclass
+++ b/classes/tdx-signed-dmverity.bbclass
@@ -1,0 +1,6 @@
+# a ramdisk is required to check the signature of the rootfs image
+INITRAMFS_IMAGE ?= "tdx-reference-ramdisk-image"
+INITRAMFS_IMAGE_BUNDLE ?= "0"
+
+# use the FIT image with the ramdisk
+KERNEL_IMAGE_NAME:prepend = "${INITRAMFS_IMAGE_NAME}-"
--- a/classes/tdxref-signed.bbclass
+++ b/classes/tdxref-signed.bbclass
@@ -1,0 +1,5 @@
+# sign bootloader and kernel artifacts
+inherit tdx-signed
+
+# boot a signed rootfs via dm-verity
+inherit tdx-signed-dmverity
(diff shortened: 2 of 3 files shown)


sergioprado:
> LGTM. Just have some questions to help me understand some points here. So for someone building the Toradex BSP reference image. They would want to inherit `tdxref-signed.bbclass` in their build, correct?

Exactly.

> In which case, is the goal to make this bbclass become the BSP reference equivalent for the `torizon-signed.bbclass` that is in `meta-toradex-torizon`? Or will it's purpose differ a bit?

Yeah, the idea is to have just two major classes: `tdxref-signed.bbclass` for BSP users and `torizon-signed.bbclass` for Torizon users.

> Also, and I might be jumping ahead. I assume it wouldn't make too much sense to inherit `torizon-signed` and `tdxref-signed` in the same build, right? Would there be any issues if someone inadvertently did that?

It doesn't make sense to inherit both, and it will probably not work, since they will use different technologies for rootfs authenticity check (dm-verity for BSP and composefs for Torizon).

jsrc27:
Okay, perfect thanks for clarifying my understanding here. Well I have no more points to raise. @rborn-tx Did you want to take a look here as well?

rborn-tx:
@jsrc27

> Did you want to take a look here as well?

Absolutely. I didn't have time yet but I intend to do it before the weekend.

rborn-tx:
The implementation so far looks good to me. Since it's been approved already, let me go ahead and merge it.

PR #12, opened 2024-03-05
Update kernel arguments to synchronize with meta-toradex-torizon

rborn-tx:
At the moment the kernel arguments expected by the "kernel command line protection" feature are kept in the present layer. Here we update them to sync with the recently merged changes in `meta-toradex-torizon`:

https://github.com/torizon/meta-toradex-torizon/pull/33


sergioprado:
Looks good to me, merging it.

PR #13, opened 2024-03-12
Boot a verified rootfs image via dm-verity

sergioprado:
This changes the rootfs to a verity image.

https://docs.kernel.org/admin-guide/device-mapper/verity.html

The verity root hash is added to the ramdisk and checked at boot time when mouting the rootfs. The rootfs will be mounted and the system will boot only if the hash in the ramdisk matches the root hash from the verity image.

Built and runtime tested on Verdin iMX8MP, Colibri iMX6, Colibri iMX7 and Verdin AM62.


sergioprado:
@rborn-tx @jsrc27 this is ready to be reviewed, but I marked as a draft because it depends on a couple of changes on the BSP side.

You can review, but please don't merge it yet.

jsrc27 (approved):
LGTM. Seems pretty straightforward, nothing much to add from my side.

rborn-tx:
LGTM as well.

sergioprado:
@rborn-tx @jsrc27 thanl you both for the review.

I will leave this as a draft until the required changes on the BSP are merged. As soon as they are merged, I will merge this one.

sergioprado:
@rborn-tx @jsrc27 Since you both already reviewed this, I am merging now.

There were two patches pending at the BSP side.

One of them was about changing the class that generates the Tezi image. This one was merged. See https://git.toradex.com/cgit/meta-toradex-bsp-common.git/commit/?h=kirkstone-6.x.y&id=0513615cb0383d7ac0695e865884897493339d62.

The other one was about adding two layers (`meta-security` and `meta-perl`) to the default `bblayers.conf` file from the BSP. This one was rejected (but still under discussion).

For now, I will just add to the documentation the information about the dependencies. I am working on the documentation and will submit another PR soon.

PR #14, opened 2024-03-21
Document rootfs signing support with dm-verity

sergioprado:
I am also adding the layer dependencies in the `layer.conf` file and documenting them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,16 +6,38 @@ For more information on available Toradex SoMs please visit:
 
 https://www.toradex.com/computer-on-modules
 
+# Layer dependencies
+
+This layer depends on:
+
+```
+URI: git://git.openembedded.org/openembedded-core
+layers: meta
+branch: kirkstone
+revision: HEAD
+
+URI: git://git.yoctoproject.org/meta-security
+branch: kirkstone
+revision: HEAD
+
+URI: git://git.toradex.com/meta-toradex-nxp.git
+branch: kirkstone
+revision: HEAD
+
+URI: git://git.toradex.com/meta-toradex-ti.git
+branch: kirkstone
+revision: HEAD
+```
+
 # Supported features and SoMs
 
 This layer supports the following security features:
 
 - Secure boot
-
-The level of support of the above features is:
-
-- BSP reference images: tested during development but not in Toradex's CI environment, integration effort is expected.
-- Torizon OS (formerly named TorizonCore): tested and integrated.
+  - Bootloader signature checking.
+  - U-Boot hardening.
+  - FIT image signature checking (kernel, DTBs and ramdisk).
+  - Rootfs signature checking via `dm-verity`.
 
 The features are currently supported on the following SoMs:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -6,16 +6,37 @@ For more information on available Toradex SoMs please visit:
...
 
 - Secure boot
-
-The level of support of the above features is:
-
-- BSP reference images: tested during development but not in Toradex's CI environment, integration effort is expected.
-- Torizon OS (formerly named TorizonCore): tested and integrated.
+  - Bootloader signature checking.
+  - FIT image signature checking (kernel, DTBs and ramdisk).
+  - Rootfs signature checking via `dm-verity`.
```
jsrc27, line 39: Do you think we should specify the `dm-verity` support is only meant for tdxref in mind?

jsrc27 (commented):
For the most part looks okay to me.

README.md
sergioprado: Yeah, I thought about this. But I decided to not do it mainly because this layer is a generic BSP layer (I even removed references to Torizon OS in the documentation).

I have done this because @rborn-tx decided to implement any Torizon specific logic for the secure boot in the meta-toradex-torizon layer.

So now meta-toradex-torizon is just a user of this layer, and there is no specific Torizon metadata here. Anyway, since a "Torizon user" might be interested in not using dm-verity, I documented how to disable it in the "Configuring rootfs image signing" session.

@rborn-tx do you think this approach makes sense?

jsrc27: Ahh okay I see the logic now. My worry was that a user would try to use dm-verity on Torizon OS (or any OSTree enabled OS). Despite OSTree and dm-verity not being very practical together.

rborn-tx: > do you think this approach makes sense?

@sergioprado Yeah, I believe it does.

However, thinking about the current state of the `torizon-signed` class (on `meta-toradex-torizon`) which is still not ready to use and the fact that the main documentation page talking about [Secure Boot on Torizon OS](https://developer.toradex.com/torizon/security/secure-boot-on-torizoncore/) is referencing `tdx-signed` as the class to be used by Torizon OS users wanting secure boot, my feeling is that we shouldn't hide `tdx-signed` too much because it would probably be a bit confusing to users getting here from the said documentation.

Considering that, my suggestion would be for us to explicitly say that we have two classes and state what each one of them covers. An idea would be to reorganize the "Secure Boot" section a bit differently, more or less like this:

```
The present layer offers the following features related to secure boot (details on the level of support and how to configure might vary depending on the SoM):

- Bootloader signature checking: ...
- U-Boot hardening: ...
- FIT image signature checking: ...
- Rootfs signature checking: ...

These features are described with more detail in the following sessions. The features actually enabled depend on the class chosen; the options are:

- `tdxref-signed`: all features listed above are enabled.
- `tdx-signed`: all features are enabled except the "Rootfs signature checking"; this class may be more suitable than `tdxref-signed` if the rootfs protection via `dm-verity` is not desired (for example, because it's not compatible with the used distro or because the rootfs is protected by other means).

To use the selected class, inherit from it globally by adding the following to an OE configuration file (e.g. your `local.conf`):

\```
INHERIT += "<chosen-class>"
\```
```

With this, the sentence about `tdx-signed` in section "Rootfs signature checking" may no longer be needed, even though I see no problem in keeping it.

Notice above that I wrote that "all features are enabled" when tdxref-signed is used; is this true? I mean, is also the hardening enabled?

sergioprado: > Considering that, my suggestion would be for us to explicitly say that we have two classes and state what each one of them covers. An idea would be to reorganize the "Secure Boot" section a bit differently, more or less like this:

@rborn-tx thanks a lot for the review. I think it makes sense. I have done the changes and updated the PR. This is ready for review.

> Notice above that I wrote that "all features are enabled" when tdxref-signed is used; is this true? I mean, is also the hardening enabled?

It depends on the SoM. AFAIK, currently the only not implemented feature is U-Boot hardening support on Verdin AM62. I added a note to make this clear.

rborn-tx: @sergioprado Thanks for the changes. They look good to me. I'm approving the PR.
@jsrc27 Do you want to take a look before we merge this?

rborn-tx:
Approved!

rborn-tx (approved):
Approved!

jsrc27 (approved):

README.md
jsrc27: Looks all good to me as well, I'll just go ahead and click merge then.

PR #18, opened 2024-04-05
Introduce encryption support

sergioprado:
This PR introduces encryption support for Toradex modules.

Currently, it only supports the `cleartext` backend for storing the encryption key. Support for other backends (CAAM, TPM, OP-TEE) will be introduced in future PRs.

Tested on Verdin iMX8MP.


classes/tdx-encrypted.bbclass
```diff
@@ -0,0 +1,39 @@
+# Enable encryption support
+TDX_ENC_ENABLE = "1"
+
+# Encryption key backend
+# This variable defines how the encryption key is managed
+# Available options:
+#    cleartext -> key is stored in clear text (for testing purposes only!)
+TDX_ENC_KEY_BACKEND ?= ""
```
jsrc27, line 8: Is the intent here to not put a default so that users must consciously make a decision on the configuration?

sergioprado: Exactly. I don't think we should leave the cleartext backend enabled by default because it is insecure.

But as soon as we have the CAAM backend implemented, then I think we can set it by default in iMX based modules.

I also plan to write some documentation after implementing CAAM support.

jsrc27: Makes sense to me.

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
```diff
@@ -0,0 +1,31 @@
...
+
+SYSTEMD_SERVICE:${PN} = "tdx-enc-handler.service"
+
+do_install() {
+    install -d ${D}${sbindir}
+    install -m 0755 ${WORKDIR}/tdx-enc-${TDX_ENC_KEY_BACKEND}.sh ${D}${sbindir}/tdx-enc.sh
+
+    sed -i 's|###TDX_ENC_KEY_FILE###|${TDX_ENC_KEY_FILE}|g' ${D}${sbindir}/tdx-enc.sh
+    sed -i 's|###TDX_ENC_STORAGE_LOCATION###|${TDX_ENC_STORAGE_LOCATION}|g' ${D}${sbindir}/tdx-enc.sh
+    sed -i 's|###TDX_ENC_STORAGE_MOUNTPOINT###|${TDX_ENC_STORAGE_MOUNTPOINT}|g' ${D}${sbindir}/tdx-enc.sh
```
rborn-tx, line 27: Just to follow a pattern in this layer, how about using `@@<variable-to-be-replaced>@@` here? See https://github.com/toradex/meta-toradex-security/blob/kirkstone-6.x.y/recipes-bsp/imx-mkimage/files/mx8_sign.sh#L109.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-handler.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Encryption handler for Toradex modules
+After=systemd-udev-settle.service
+Wants=systemd-udev-settle.service
```
rborn-tx, line 4: Perhaps we need a `Before` here too since the service is going to mount something that might be needed by other services/applications (including Docker ones)? Apparently the normal case for local filesystems is to add a `Before=local-fs.target`. Btw, `containerd` has an `After=local-fs.target` which should allow apps to run when required storage is fully mounted...

rborn-tx (commented):
Raised a couple points to consider before merging...

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-handler.service
rborn-tx: I forgot to mention that the service here not only mounts the encrypted partition but also formats it when needed which could be time consuming; that I believe makes the dependency relationships even more important to avoid timing issues...

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
sergioprado: Oh yeah, I forgot about this. I will follow the same patter then. Thanks for pointing that out.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-handler.service
sergioprado: Hmmmm, yeah, `local-fs.target` is indeed a much better target to depend on. I will change that.

sergioprado:
@rborn-tx @jsrc27 PR updated with the following changes:

- Using `@@` for string substitution.
- `tdx-enc-handler.service` configured to be executed before `local-fs.target`.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-handler.service
rborn-tx: @sergioprado 

> tdx-enc-handler.service configured to be executed before local-fs.target.

I noticed you also removed the `After=systemd-udev-settle.service` dependency. Is this correct?

For me the dependency on `systemd-udev-settle.service` seemed appropriate. For example, a user could have created udev rules that they want applied/executed before anyone attempts to mount/access the disk/partition, right?

Also, they could be referencing the partition by its label or UUID rather than the not-stable kernel name, i.e. they might want to set `TDX_ENC_STORAGE_LOCATION = "/dev/disk/by-label/<PARTLABEL>"`, but the latter symlink would only exist after event processing is done by udev.

sergioprado:
@rborn-tx just following up on this PR. Are you still reviewing it?

rborn-tx:
> @rborn-tx just following up on this PR. Are you still reviewing it?

@sergioprado Have you seen my last comment https://github.com/toradex/meta-toradex-security/pull/18#discussion_r1564121053? Maybe you resolved that thread by mistake?

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-handler.service
sergioprado: @rborn-tx I tried to leave the `After=systemd-udev-settle.service` there, but it causes circular dependencies when used in conjunction with `Before=local-fs.target`.

```
# journalctl -u tdx-enc-handler
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Found ordering cycle on systemd-journal-catalog-update.service/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Found dependency on local-fs.target/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Found dependency on tdx-enc-handler.service/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Found dependency on sysinit.target/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Job systemd-journal-catalog-update.service/start deleted to break ordering cycle starting with sysinit.target/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Found ordering cycle on local-fs.target/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Found dependency on tdx-enc-handler.service/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Found dependency on sysinit.target/start
Jul 27 12:20:04 localhost systemd[1]: sysinit.target: Job local-fs.target/start deleted to break ordering cycle starting with sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found ordering cycle on systemd-journal-catalog-update.service/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on local-fs.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on tdx-enc-handler.service/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Job systemd-journal-catalog-update.service/start deleted to break ordering cycle starting with sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found ordering cycle on local-fs.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on tdx-enc-handler.service/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Job local-fs.target/start deleted to break ordering cycle starting with sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found ordering cycle on systemd-journal-catalog-update.service/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on local-fs.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on tdx-enc-handler.service/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Job systemd-journal-catalog-update.service/start deleted to break ordering cycle starting with sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found ordering cycle on local-fs.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on tdx-enc-handler.service/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Found dependency on sysinit.target/start
Jul 27 12:20:05 localhost systemd[1]: sysinit.target: Job local-fs.target/start deleted to break ordering cycle starting with sysinit.target/start
Jul 27 12:20:06 localhost systemd[1]: Starting Encryption handler for Toradex modules...
Jul 27 12:20:06 localhost tdx-enc.sh[980]: cleartext: Generating encryption key...
Jul 27 12:20:06 localhost tdx-enc.sh[980]: cleartext: Setting up partition with LUKS...
Jul 27 12:20:06 localhost tdx-enc.sh[1023]: Device /dev/sda1 is not a valid LUKS device.
Jul 27 12:20:06 localhost tdx-enc.sh[980]: cleartext: Formatting partition with LUKS...
Jul 27 12:20:13 verdin-imx8mp-15139757 tdx-enc.sh[980]: cleartext: Mounting encrypted partition...
Jul 27 12:20:13 verdin-imx8mp-15139757 tdx-enc.sh[980]: cleartext: Formatting encrypted partition with ext4...
Jul 27 12:20:38 verdin-imx8mp-15139757 tdx-enc.sh[980]: cleartext: Success!
Jul 27 12:20:38 verdin-imx8mp-15139757 systemd[1]: Finished Encryption handler for Toradex modules.
```

I dig deeper into this, and it seems depending on `systemd-udev-settle.service` is not really recommended. From its man page:

```
$ man systemd-udev-settle.service
...
Using this service is not recommended. There can be no guarantee that hardware is fully discovered at any specific
time, because the kernel does hardware detection asynchronously, and certain buses and devices take a very long
time to become ready, and also additional hardware may be plugged in at any time. Instead, services should
subscribe to udev events and react to any new hardware as it is discovered. Services that, based on configuration,
expect certain devices to appear, may warn or report failure after a timeout. This timeout should be tailored to the
hardware type. Waiting for systemd-udev-settle.service usually slows boot significantly, because it means waiting for
all unrelated events too.
...
```

Also, it was deprecated a while ago and might be removed from systemd anytime. See https://github.com/systemd/systemd/blob/main/src/udev/udevadm-settle.c#L138.

For now, I don't think we can provide a generic solution here. If a BSP user has any specific needs, she would need to tweak the service for his own needs.

What do you think? Do you see any other solution?

rborn-tx: @sergioprado I was curious and did a quick test by adding a `DefaultDependencies=no` to the Unit section of your .service file which solved the dependency loop. So, this could be a solution, but given the recommendation you mentioned it seems it's best to just avoid the dependency on `systemd-udev-settle` altogether.

I think the need for disabling the default dependencies comes from:

```
$ man systemd.special
...
basic.target
           A special target unit covering basic boot-up.

           systemd automatically adds dependency of the type After= for this target
           unit to all services (except for those with DefaultDependencies=no).

           Usually, this should pull-in all local mount points  <=====
           plus /var, /tmp and /var/tmp, swap devices, sockets, timers, path units
           and other basic initialization necessary for general purpose daemons.
...
```

Now, removing the dep with `systemd-udev-settle` opens the possibility of the block device defined by `TDX_ENC_STORAGE_LOCATION` not being available when the service is started which is not good. Considering that the device is user-configurable, perhaps we should add some comment saying that the service **will likely** need to be tweaked to ensure the presence of the device. As we don't have documentation on this feature yet, we could for example add such a comment to the description of `TDX_ENC_STORAGE_LOCATION`. What do you think?

In the example I gave where a user wanted to define `TDX_ENC_STORAGE_LOCATION` by a device label it seems it would require adding this to the service (that could be done by an override):

```
[Unit]
Wants=dev-disk-by\x2dlabel-[SOME-LABEL].device
After=dev-disk-by\x2dlabel-[SOME-LABEL].device
```

Suggestion: Maybe this idea could also be part of the comment since I think it would be a good practice that most users wouldn't do by themselves unless they have a heads up; the `Wants` might be optional depending on user needs.

Anyways, I tried the above override and it also caused a dependency loop! But that was again solved by disabling the default dependencies. Thus, I think we should already add `DefaultDependencies=no` to the `tdx-enc-handler.service` file. Makes sense?

rborn-tx: @sergioprado Do you know about `blockdev@.target` (described in systemd.special manual)? It could be relevant to our use case here.

rborn-tx: @sergioprado 

> oh man, can we revert to sysvinit scripts? :-)

:rofl: 

Well most of what I said was related to documentation, but leaving any documentation stuff for the future sounds okay to me.

The only remaining topic on the implementation side is regarding the setting of `DefaultDependencies`. You mentioned:

> I guess this is just masking the issue, and not really solving it, right?

As per the documentation I referenced before, systemd will add a dependency on `basic.target` to all services (if they don't have `DefaultDependencies=no`) which in turn will pull in all local mounts as dependencies. Since the encryption service being introduced here is a provider of an encrypted block device with a corresponding (local) mount, my first impression was that it shouldn't itself depend on `basic.target`.

Having said that, as you tested it and the presence of the dependency does not seem to cause any visible issues, I'm okay with moving ahead with the current implementation unchanged. Approving the PR...

rborn-tx (approved):

PR #19, opened 2024-04-18
Introduce CAAM backend support for iMX based modules

sergioprado:
This PR introduces CAAM backend support for iMX based modules.

The current implementation only works on SoMs based on the downstream kernel (e.g. iMX8). On iMX6 and iMX7 using the upstream kernel, despite the CAAM driver being enabled, the `caam-keygen` utility cannot find the CAAM interface:

```
# caam-keygen create dek ccm -s 32
Unable to open device /dev/caam-keygen

# dmesg | grep caam
[    1.636645] caam 2100000.crypto: device ID = 0x0a16010000000100 (Era 4)
[    1.643337] caam 2100000.crypto: job rings = 2, qi = 0
[    1.651916] caam algorithms registered in /proc/crypto
[    1.657379] caam 2100000.crypto: registering rng-caam
[    1.662731] caam 2100000.crypto: rng crypto API alg registered prng-caam
```

I will investigate that in the following days.

Anyway, I expect the implementation to support iMX6/iMX7 using the upstream kernel to be transparent, so that means this PR is ready to be reviewed/merged.

The implementation was tested on Verdin iMX8MP and Apalis iMX8.


jsrc27 (approved):
Seems fairly straightforward to me.

MMG-DG:
For what it is worth...
I can verify this works for my builds using the tdxref image manifest.

The only possible issue I have experienced is 1 in 20 times or so the encrypted partition can get into a corrupted state. I believe if something interferes with the encryption process then the content of the encrypted partition seems to end up getting wiped. However, this I believe is an issue with one of our other layers, unless anyone can say otherwise.

sergioprado:
> The only possible issue I have experienced is 1 in 20 times or so the encrypted partition can get into a corrupted state. I believe if something interferes with the encryption process then the content of the encrypted partition seems to end up getting wiped.

@MMG-DG thanks for reporting. I didn't experience that in my tests. If that happens again, would you please collect and provide us the logs?

```
# systemctl status tdx-enc-handler
# journalctl -u tdx-enc-handler
```

rborn-tx (approved):
Implementation looks good to me.

@sergioprado I don't know if you want to wait @EstebanSannin feedback here. If not, feel free to merge it.

EstebanSannin:
Looks good to me.

EstebanSannin (approved):
Looks good to me.

sergioprado:
Thank you all for reviewing! I am merging this PR.

Issue #20, opened 2024-04-21
U-Boot hardening doesn't work on BSP images


MMG-DG:
There seems to be some issues with the set of whitelisted commands.

I am using an im8xmp verdin SoM.
My build is based on the tdxref default manifest configuration (not the TorizonOS one).
I have added the `INHERT = "tdx-signed"` and other variables to use my custom certs and keys.

However, after closing the device I am unable to boot it.
I have now got a patch that sets my open allowed list to the same as the closed list so I can more easily debug and still flash new images using TEZI.

What I have found out is that my boot_cmd to load from mmc includes the use of `part uuid` which is only listed as `CMD_CAT_PART_READ`.
I believe [0001-toradex-common-add-command-whitelisting-modules.patch](https://github.com/toradex/meta-toradex-security/blob/kirkstone-6.x.y/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch#L602) should include the `CMD_CAT_NEEDED` category so that it is an allowed command during boot.
Changing this one value appears to work for me, but not sure how to prepare a change and PR for this fix (I am a noob at yocto things).

Here is essentially what is being run to load the mmc2 root partition:
```
env set uuid_set 'part uuid ${root_devtype} ${root_devnum}:${root_part} uuid'
env set rootfsargs_set 'run uuid_set && env set rootfsargs root=PARTUUID=${uuid} ro rootwait'
```

sergioprado:
@MMG-DG thanks for reporting and debugging the issue.

Maybe some recent change in the Toradex BSP caused this. Your proposed fix makes sense to me.

@rborn-tx what do you think?

rborn-tx:
@sergioprado Have we tested booting tdxref with hardening enabled? If so, then we should have seen messages saying that some command didn't pass the whitelist (which would prevent the boot when the device was closed). From the warnings, we should evaluate whether or not the commands are **always safe** and, if so, add them to the whitelist. This would require code changes so it would be normally be done by us at Toradex.

Users who decide to customize the boot script shouldn't need to do something like that; instead they would configure the set of allowed commands via the control FDT even on a compiled image (but we don't have the tooling for that yet and that's why this is not yet documented).

> Maybe some recent change in the Toradex BSP caused this.

Checking the tdxref boot script I don't see recent changes. Thus my feeling is that we may have overlooked warning messages when testing tdxref.

rborn-tx:
[...]

PR #22, opened 2024-04-23
tdx-signed: enable hardening only on Torizon OS images

rborn-tx:
Since hardening is supported and tested only on Torizon OS, enable it by default only when building such images. At the moment hardening features are not expected to work out of the box with BSP images, so keeping it enabled with them will cause more harm than good (Since hardening is supported and tested only on Torizon OS, enable it by default only when building such OS images. At the moment the hardening features are not expected to work out of the box with BSP images, so keeping it enabled will cause more harm than good (as it was the case on issue #20).

### DIFF
--- a/classes/tdx-signed-harden.inc
+++ b/classes/tdx-signed-harden.inc
@@ -9,10 +9,23 @@
 # combination UBOOT_SIGN_ENABLE="0" TDX_UBOOT_HARDENING_ENABLE="1" is not very
 # helpful and can prevent the loading of non-FIT images.
 #
-# Because of the above, we enable the hardening only if both TDX_IMX_HAB_ENABLE
-# and UBOOT_SIGN_ENABLE are set, by default.
+# Moreover, currently the hardening is tested only with Torizon OS and it's
+# not expected to work out of the box with BSP images.
 #
-TDX_UBOOT_HARDENING_ENABLE ?= "${@'1' if bb.utils.to_boolean(d.getVar('TDX_IMX_HAB_ENABLE')) and bb.utils.to_boolean(d.getVar('UBOOT_SIGN_ENABLE')) else '0'}"
+# Because of the above, we enable the hardening only when building Torizon OS
+# images provided that both TDX_IMX_HAB_ENABLE and UBOOT_SIGN_ENABLE are set, by
+# default.
+#
+def default_tdx_uboot_hardening_enable(d):
+    # TODO: Remove this condition once support for BSP images is added.
+    if "torizon" not in d.getVar("OVERRIDES").split(":"):
+        return False
+    if (bb.utils.to_boolean(d.getVar('TDX_IMX_HAB_ENABLE')) and
+        bb.utils.to_boolean(d.getVar('UBOOT_SIGN_ENABLE'))):
+        return True
+    return False
+
+TDX_UBOOT_HARDENING_ENABLE ?= "${@'1' if default_tdx_uboot_hardening_enable(d) else '0'}"
 
 # Configure one of the hardening features: the "bootargs protection"; with this
 # protection, the fixed part of the kernel command line (boot arguments) are
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ The hardening features above are controlled by the following variables:
 
 | Variable | Description | Default value |
 | :------- | :---------- | :------------ |
-| `TDX_UBOOT_HARDENING_ENABLE` | Enable hardening features as a whole | `1` if both `TDX_IMX_HAB_ENABLE` and `UBOOT_SIGN_ENABLE` are set or 0 otherwise |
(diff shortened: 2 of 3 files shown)


classes/tdx-signed-harden.inc
```diff
@@ -9,10 +9,23 @@
...
+def default_tdx_uboot_hardening_enable(d):
+    # TODO: Remove this condition once support for BSP images is added.
+    if "torizon" not in d.getVar("OVERRIDES").split(":"):
+        return False
+    if (bb.utils.to_boolean(d.getVar('TDX_IMX_HAB_ENABLE')) and
+        bb.utils.to_boolean(d.getVar('UBOOT_SIGN_ENABLE'))):
+        return True
+    return False
+
+TDX_UBOOT_HARDENING_ENABLE ?= "${@'1' if default_tdx_uboot_hardening_enable(d) else '0'}"
```
jsrc27, line 28: I suppose someone could still overwrite this variable in a conf file somewhere to be set for non-Torizon OS builds right? Should we try and print a warning in such cases or just leave the user to the consequences?

I'm fine with either honestly since the user would be making a conscious decision despite our documentation.

rborn-tx: Good point. I've added the check to stop build if one tries to use the unsupported configuration.

rborn-tx: Btw, I chose to throw a fatal error because we were already doing that for the other combination of unsupported configuration. Opinions?

sergioprado:
Ideally, I would expect this layer to be Torizon agnostic (not Torizon logic here). But I understand this is a workaround to fix the current situation.

The changes look good to me. I will wait for @jsrc27 final revision before merging this.

classes/tdx-signed-harden.inc
MMG-DG: Is this to say that it will never let someone build in an unsupported configuration?
I would suggest a warning instead as it may be wanted regardless of support from toradex (e.g. custom support is being added by the implementer).

jsrc27: > Btw, I chose to throw a fatal error because we were already doing that for the other combination of unsupported configuration. Opinions?

I don't have a super strong opinion on fatal error vs warning. Though I do admit that throwing a fatal error more clearly communicates our intent and expected usage of this layer, and is consistent with other cases.

> I would suggest a warning instead as it may be wanted regardless of support from toradex (e.g. custom support is being added by the implementer).

Could you elaborate? If a contributor adds support and submits a PR here then we could easily revise the error/warning handling in this case at that time. If experimenting then simply removing/commenting out the point of throwing the error would be simple as well.

MMG-DG: The problem would be that another implementer would not be able to build or test as the bbfatal would stop the build process.
The likely hood is that I will be looking into this in the next few weeks to implement any required changes in one of our layers.
Once tested and working, I would then apply the same changes and feed back as a PR here.
My main problem is that my main development machine is offline with a nightly update (fork) of the repos we depend on.

jsrc27: > The problem would be that another implementer would not be able to build or test as the bbfatal would stop the build process.

I don't see this as a problem. The implementer in question would just change the bbfatal in their implementation then and submit that as part of the PR. This does not permanently prevent anything in the future. This is just about communicating the intent and current state of the layer as it exists now.

MMG-DG: Just to clarify my previous....
The build environment I work in does not allow me direct access to modify any public sources.
Ideally, I would pull the git repo, figure out what changes are needed to fix any warnings and issues, then move over to my online system and "eyeball" the changes over into a writable web-based version that I would then create a PR for.

MMG-DG: If there was maybe an extra "undocumented" flag that I could use to "ignore" the error, then that would also be OK. As I could then set that while developing and fixes.
Another "Hack" would be to add the checked string into OVERRIDES, but not sure that would be a good idea as not sure if that may trigger other changes in dependant layers.

rborn-tx: @MMG-DG 

> Another "Hack" would be to add the checked string into OVERRIDES, but not sure that would be a good idea as not sure if that may trigger other changes in dependant layers.

This should work for your use case since the only other layer depending on the "torizon" override is `meta-toradex-torizon` which you're not supposed to be using.

sergioprado: @MMG-DG we appreciate a lot your feedback.

We will merge this because having a feature that is currently broken might cause more harm than good.

Be aware that this PR is just a workaround. We plan to review this soon, so it is possible that this feature will be available to BSP users soon.

For your specific used case, in case you want to take the risks, please add `torizon` to `OVERRIDES` to use the U-Boot hardening feature.

PR #24, opened 2024-05-10
Change the CAAM backend to use the kernel Trusted Keys infrastructure

sergioprado:
Trusted keys is a feature that makes it possible to have a trusted back-end for generating and managing keys in the kernel key ring service.

https://docs.kernel.org/security/keys/trusted-encrypted.html

Since CAAM is already one of the supported root of trust for key storage, we can leverage it.

Build and runtime tested on Colibri iMX6, Colibri iMX7 and Verdin iMX8MP.


jsrc27 (approved):
Overall looks fine to me, just some minor non-important comments.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-handler.service
```diff
@@ -1,6 +1,8 @@
 [Unit]
 Description=Encryption handler for Toradex modules
+DefaultDependencies=no
```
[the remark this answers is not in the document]
sergioprado: For the default BSP image, it works without this line, so it is just an optimization. But since the final user might have early-boot services that might depend on the encrypted data partition, the soon we run this, the better.

Looking at the default early-boot systemd services, most of them (more than 60) set this option, so I thought we should do the same:

```
# grep -R DefaultDependencies=no /lib/systemd/system/*.service | wc -l
66

# grep -R DefaultDependencies=no /lib/systemd/system/*.service
/lib/systemd/system/debug-shell.service:DefaultDependencies=no  
/lib/systemd/system/emergency.service:DefaultDependencies=no          
/lib/systemd/system/initrd-cleanup.service:DefaultDependencies=no              
/lib/systemd/system/initrd-parse-etc.service:DefaultDependencies=no   
/lib/systemd/system/initrd-switch-root.service:DefaultDependencies=no
/lib/systemd/system/initrd-udevadm-cleanup-db.service:DefaultDependencies=no
/lib/systemd/system/kmod-static-nodes.service:DefaultDependencies=no    
/lib/systemd/system/ldconfig.service:DefaultDependencies=no       
/lib/systemd/system/modprobe@.service:DefaultDependencies=no         
/lib/systemd/system/pam_namespace.service:DefaultDependencies=no   
/lib/systemd/system/quotaon.service:DefaultDependencies=no           
/lib/systemd/system/rescue.service:DefaultDependencies=no          
/lib/systemd/system/rpcbind.service:DefaultDependencies=no
/lib/systemd/system/system-update-cleanup.service:DefaultDependencies=no                                                                                                                                           
/lib/systemd/system/systemd-ask-password-console.service:DefaultDependencies=no
/lib/systemd/system/systemd-backlight@.service:DefaultDependencies=no
/lib/systemd/system/systemd-exit.service:DefaultDependencies=no
/lib/systemd/system/systemd-fsck-root.service:DefaultDependencies=no
/lib/systemd/system/systemd-fsck@.service:DefaultDependencies=no
/lib/systemd/system/systemd-halt.service:DefaultDependencies=no
/lib/systemd/system/systemd-hibernate-resume@.service:DefaultDependencies=no
/lib/systemd/system/systemd-hibernate.service:DefaultDependencies=no
/lib/systemd/system/systemd-hwdb-update.service:DefaultDependencies=no
/lib/systemd/system/systemd-hybrid-sleep.service:DefaultDependencies=no
/lib/systemd/system/systemd-initctl.service:DefaultDependencies=no
/lib/systemd/system/systemd-journal-catalog-update.service:DefaultDependencies=no
/lib/systemd/system/systemd-journal-flush.service:DefaultDependencies=no
/lib/systemd/system/systemd-journald.service:DefaultDependencies=no
/lib/systemd/system/systemd-kexec.service:DefaultDependencies=no
/lib/systemd/system/systemd-machine-id-commit.service:DefaultDependencies=no
/lib/systemd/system/systemd-modules-load.service:DefaultDependencies=no
/lib/systemd/system/systemd-network-generator.service:DefaultDependencies=no
/lib/systemd/system/systemd-networkd-wait-online.service:DefaultDependencies=no
/lib/systemd/system/systemd-networkd.service:DefaultDependencies=no
/lib/systemd/system/systemd-poweroff.service:DefaultDependencies=no
/lib/systemd/system/systemd-pstore.service:DefaultDependencies=no
[...]
```

Anyway, I don't have any benchmark on if setting this line makes the boot faster or not.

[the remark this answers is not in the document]
rborn-tx: To be honest, I'm not sure why a dependency loop error/warning is not triggered without `DefaultDependencies=no`, as I mentioned in the previous PR (https://github.com/toradex/meta-toradex-security/pull/18#discussion_r1569731410).

Anyways, I think disabling default dependencies is the proper setting for us.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -14,8 +14,8 @@ TDX_ENC_STORAGE_MOUNTPOINT="@@TDX_ENC_STORAGE_MOUNTPOINT@@"
 # dm-crypt device to be created
 TDX_ENC_DM_DEVICE="encdata"
 
-# CAAM black key name
-TDX_ENC_KEY_NAME="dek"
+# CAAM encrypted key name
+TDX_ENC_KEY_NAME="dek.bb"
```
rborn-tx, line 18: Is `.bb` a common extension for this usage here?

recipes-kernel/linux/linux-dm-crypt.inc
```diff
@@ -1,7 +1,4 @@
-SRC_URI += " \
+SRC_URI:append = "\
     file://dm-crypt.cfg \
-"
-
-SRC_URI:append:imx-generic-bsp = " \
-    file://caam.cfg \
+    file://${TDX_ENC_KEY_BACKEND}.cfg \
```
rborn-tx, line 3: Out of curiosity, what would happen if one tries to enable CAAM on a non-NXP device? I'd guess there would be errors during kernel build in such case my question would be: do we want to prevent them? (I'm ok either way)

rborn-tx (approved):
I raised a couple questions but the implementation looks good to me regardless of the answers (so, I'm already approving the PR).

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
sergioprado: Hmmm, I am not sure if there is a common extension for this use case. I am just following here what NXP has used in its manual.

`bb` stands for black blob, and that is how they call the blob that has the encrypted key.

Anyway, naming stuff is hard, so I am open to new ideas. :-)

recipes-kernel/linux/linux-dm-crypt.inc
sergioprado: The build will not fail. The kernel will be compiled with CAAM support (which is supported upstream), even on non-iMX devices.

The error will be at runtime. In case there is no CAAM device available, the Trusted Keys subsystem will fail to create a key backed by CAAM.

rborn-tx: Nice. So the user should be able to realize they did something wrong.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
rborn-tx: Okay. It just felt kind of funny to see the same extension used with BitBake recipes :slightly_smiling_face:

Issue #25, opened 2024-05-13
Support for "In-Place" Encryption


MMG-DG:
Howdy,

I have been playing with the latest caam encryption code in my dev environment and have noticed that the partition to be encrypted will be formatted on initial setup.

Is it (or could it be) possible to check if the partition contains data and move it before encrypting the partition and then move the data back after?
I have my build configured to put data into my "soon-to-be-encrypted" partition, but then once it is mounted it is empty.
It would be nice if the encryption setup accounted for any data on the partition before potentially wiping it.

Not sure if dmsetup has the ability to encrypt the partition with data in-place, or if it would be possible to "detect" any data and move to a RAM-Drive while setting up the encrypted partition, and then move back once setup is completed?

MMG-DG:
Have made a quick fix for this issue.
Please ignore the first linked changes (I am on windows and managed to mess up the commit).

sergioprado:
Interesting. I am not sure if this can be done transparently via dmsetup, but can certainly be done by the script that handles the encryption.

I linked your PR (https://github.com/toradex/meta-toradex-security/pull/26) to this issue.

MMG-DG:
@sergioprado - Can you unlink the old PR and link the new one #27 to this now I have updated it correctly this time.

PR #27, opened 2024-05-15
Preserve data on initial setup of encrypted partition

MMG-DG:
Added methods to backup original data to RAM Disk and then restore after partition is encrypted and formatted.


MMG-DG:
Relates to #25

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +61,20 @@ tdx_enc_key_gen() {
     fi
 }
 
+# backup original data in partition (if not encrypted)
+tdx_enc_backup_data() {
+    mkdir -p "${TDX_ENC_STORAGE_MOUNTPOINT}"
+    if mount -t ext4 ${TDX_ENC_STORAGE_LOCATION} "${TDX_ENC_STORAGE_MOUNTPOINT}"; then
```
sergioprado, line 67: I think we should have a variable (i.e. `TDX_ENC_KEEP_DATA`) so the user can enable or disable this behaviour, since some users might just want to ignore the content of the partition and format it.

This variable might be `0` by default, but users could enable this behaviour by setting it to `1`.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +61,20 @@ tdx_enc_key_gen() {
...
+        tdx_enc_log "Backing up original content..."
+        
+        mkdir /tmp/${TDX_ENC_DM_DEVICE}
+        chmod 777 /tmp/${TDX_ENC_DM_DEVICE}
```
sergioprado, line 71: Why `chmod 777` here?

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +61,20 @@ tdx_enc_key_gen() {
...
+        mount -t tmpfs -o size=$(blockdev --getsz ${TDX_ENC_STORAGE_LOCATION}) ${TDX_ENC_DM_DEVICE} /tmp/${TDX_ENC_DM_DEVICE}
```
sergioprado, line 72: Hmm... If we have 500GB of data in the partition, then we will use 500MB of RAM! That might not be desirable and can cause problems. I think we should set some limits here.

Maybe limit the ammount of data to be copied and failed if it passes a threshold? 

@rborn-tx any ideas?

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: Already on it, my most recent changes locally has `TDX_ENC_PRESERVE_DATA` but that has a bit of a ripple effect in the layer as more than 1 file needs to change for that.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +61,20 @@ tdx_enc_key_gen() {
     fi
 }
 
+# backup original data in partition (if not encrypted)
+tdx_enc_backup_data() {
+    mkdir -p "${TDX_ENC_STORAGE_MOUNTPOINT}"
+    if mount -t ext4 ${TDX_ENC_STORAGE_LOCATION} "${TDX_ENC_STORAGE_MOUNTPOINT}"; then
+        tdx_enc_log "Backing up original content..."
+        
+        mkdir /tmp/${TDX_ENC_DM_DEVICE}
```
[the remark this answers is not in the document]
MMG-DG: Will add this to my latest local changes and test today.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: No reason, this was originall copied from elsewhere.
Have removed this in my local changes and will test it today.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: Is there an easy way to determine available RAM? if so I could add a check that this size is not more than 50-75% of available RAM and fail otherwise?
That way it would be generic for any given module.

sergioprado: Interesting idea. Please limit to 30% of available memory.

This command will return free memory in kB:

```
# grep MemFree: /proc/meminfo | tr -s ' ' | cut -d ' ' -f 2
```

sergioprado: That looks like a better approach indeed! Combined with making sure we don't fill up the RAM by limiting the maximum size of the tar file.

MMG-DG: > Both the BSP images and Torizon OS have /tmp as tmpfs, so my question is: why do we need this other tmpfs mount?
> 
> I notice the size of the tmpfs here is made to match the size of the whole block device whose data is to be saved which could be a waste, right? Perhaps a better approach would be to tar up the data to be saved into a simple file in /tmp. We could limit the size of the compressed file on-the-fly by doing something like this:
> 
> ```
> tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -cj . | dd bs=1024 count=123 of=/tmp/saved.tar.bz2
> ```
> 
> where 123 is the size limit in kB (30% of the RAM as suggested).
> 
> With this, the amount of RAM temporarily used would be just the size of the data being saved after compression.
> 
> Wouldn't this work? Any downsides that you guys see?

I can certainly add the compressed tar.
With the tmpfs stuff.... (I copied code from somewhere else), I assume this is in RAM? but would it be mounted by the time this script executes?
I was also toying with the idea to make the percentage configurable (just in case) and could set the default to 30.

I will get these added to my local changes on the train home and give them a test on my image tomorrow and report back.

rborn-tx: > With the tmpfs stuff.... (I copied code from somewhere else), I assume this is in RAM?

It is.

> ...but would it be mounted by the time this script executes?

Good point, since the script is now touching `/tmp` (independent of the change I proposed) and we set `DefaultDependencies=no` I think the service should now explicitly require "tmp.mount" as a "Wants" dependency.

@sergioprado Do you agree?

thepinkmile: Initial testing looks good for these. Just looking at adding the tar bit and checking the tmpfs wants change. Should have a new commit on this later today.

MMG-DG: > Good point, since the script is now touching `/tmp` (independent of the change I proposed) and we set `DefaultDependencies=no` I think the service should now explicitly require "tmp.mount" as a "Wants" dependency.

I will add this to my local changes and see what happens. I assume this means I won't have to manually mount and create the tmpfs anymore too?

MMG-DG: One other question I have....
If this is now using the already mounted tempfs, do we need to have any error message/validation on the size of the backup?
I expect the tar commend mentioned would error if the size is greater than that?

sergioprado: > Good point, since the script is now touching /tmp (independent of the change I proposed) and we set DefaultDependencies=no I think the service should now explicitly require "tmp.mount" as a "Wants" dependency. Do you agree?

Makes sense, though I would recommend checking systemd logs to make sure we don't have any cyclic dependencies.

> I assume this means I won't have to manually mount and create the tmpfs anymore too?

Exactly, /tmp should already be mounted with tmpfs. You don't have to mount it or unmounted it.

> If this is now using the already mounted tempfs, do we need to have any error message/validation on the size of the backup?

As a general rule, yes, show error messages before returning an error from the script.

> I expect the tar commend mentioned would error if the size is greater than that?

Yes!

MMG-DG: As it turns out, that command seems to fail for me. no option '-j' apparently. I have an alternative command that seems to work.
Also, I have a possibly better solution to check if the data is too big (assuming the tmpfs is already mounted and configured correctly).
Just running another test now and will push my changes if it works.

MMG-DG:
All my latest changes have now been applied and tested on my local setup (I only have verdin imx8mp available to test with).

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
rborn-tx: @MMG-DG 

> As it turns out, that command seems to fail for me. no option '-j' apparently.

That's odd. If I run:

```
root@colibri-imx6-10866313:~# tar
BusyBox v1.35.0 () multi-call binary.

Usage: tar c|x|t [-ZzJjahmvokO] [-f TARFILE] [-C DIR] [-T FILE] [-X FILE] [LONGOPT]... [FILE]...
```

And a test seem to work.

> In my testing the command above didn't error.

I guess you're talking about the pipeline involving `dd`. That's true and reason is that when we have a shell pipeline like `a | b | c` the exit code is by default set to the last command in the pipeline; in our case the last command is `dd` which didn't fail but rather caused the failed to the preceding commands.

In bash we would solve this by simply doing `set -o pipefail` but since we're using `sh`, we need an alternative implementation such as:

```
msgs="$({ { tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -cj . || echo "ERROR" >&2; } | dd bs=1024 count=100 of=/tmp/saved.tar.bz2; } 2>&1)"
if echo "${msgs}" | grep -q ERROR; then
    # Some error happened.
fi
```

If the tar `-j` option doesn't really work, the pipeline above becomes:
```
msgs="$({ { tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -c . || echo "ERROR" >&2; } | { bzip2 -cz || echo "ERROR" >&2; } | dd bs=1024 count=100 of=/tmp/saved.tar.bz2; } 2>&1)"
```

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +67,27 @@ tdx_enc_key_gen() {
     fi
 }
 
+# backup original data in partition (if not encrypted)
+tdx_enc_backup_data() {
+    if [ ${TDX_ENC_PRESERVE_DATA} -eq 1 ]; then
+        mkdir -p "${TDX_ENC_STORAGE_MOUNTPOINT}"
+        if mount -t ext4 ${TDX_ENC_STORAGE_LOCATION} "${TDX_ENC_STORAGE_MOUNTPOINT}"; then
```
rborn-tx, line 74: I'm not sure specifying a fs type here is good. Why not leave it to auto-detection?

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +67,27 @@ tdx_enc_key_gen() {
...
+            tdx_enc_log "Backing up original content..."
+
+            TDX_ENC_PRESERVE_LOCATION=$(mktemp --suffix .tar.bz2)
```
rborn-tx, line 77: The busybox version of `mktemp` does not seem to have this `--suffix` option. It's best not to rely on it; there's no need to use a file name extension anyways.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -94,6 +121,20 @@ tdx_enc_partition_mount() {
...
 
+# restore data if available
+tdx_enc_restore_data() {
+    if [ ${TDX_ENC_PRESERVE_DATA} -eq 1 ]; then
+        tdx_enc_log "Checking backup file: ${TDX_ENC_PRESERVE_LOCATION}"
+
+        if [ -f ${TDX_ENC_PRESERVE_LOCATION} ]; then
+            tdx_enc_log "Restoring original content..."
+
+            bzip2 -cd ${TDX_ENC_PRESERVE_LOCATION} | tar --strip-components 1 -C ${TDX_ENC_STORAGE_MOUNTPOINT} -xf -
```
rborn-tx, line 132: We shouldn't need the `--strip-components` parameter here since we know the files were packed relative to the current directory (`.`).

MMG-DG: In my local testing this seemd to be required as we strip into the storage location, however the backup also included this path. So what I was seeing is data restored into `/data/data` when it should just have been `/data`.
I expect the backup function should probably be changed to avoid this instead to be more precise?

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: Will remove and test on my dev board today.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: Will remove and test on my dev board today.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: > If the tar `-j` option doesn't really work, the pipeline above becomes:

I will update my script and test locally on my dev board today and see what happens. I am wondering if there is a difference here because I am based on the tdx-reference-minimal-image? `tar` is available, `bzip2` and `gzip` are available, but tar still seems unable to use `-j` or `-z`.

Will do some more testing today and try to figure out the best option.

MMG-DG: From testing on my setup I get the following when using the `j` option:
image
Will revert to the separate bzip command and test again.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: Latest update:
Apparently I was wrong. The image I was testing was an older one. After a new clean build the backup is now correct.
I can only assume it is because the latest code now does the `dd` command instead of the `tee` I originally had.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: I believe the only thing remaining from this thread is the ability to get all errors properly. Current code still have the check, but any time I try to include the `OR`'d pipe for errors it fails.
Any ideas welcome.
I was debating about having the commands push the stderr into a file in `/tmp` and then have the check read that file?

rborn-tx: @MMG-DG Can you update the PR so we can look at the actual code?

MMG-DG: @rborn-tx - I am looking at the PR and all the code is here. Not sure why you don't see it.
I have pushed all changes to my fork's branch and it shows up here.
My guess is that because the lines have changes that this initial comment was on is not out of date it still points to the old code.

MMG-DG:
Sorry....
New line in the sand here as I had to force push my repo to get a proper rebase to happen since another PR has been closed.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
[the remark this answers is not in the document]
MMG-DG: @sergioprado - Are you happy to resolve this one as the latest code has this change.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: @sergioprado - Are you happy to resolve this one as the latest code has this change.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: @sergioprado & @rborn-tx  - Would you both be happy to resolve this comment as the latest code has all these changes in and if needed start a new conversation based on the latest code if needed?

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: @rborn-tx - Are you happy to resolve this one as the latest code has this change.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: @rborn-tx - Are you happy to resolve this one as the latest code has this change.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: @rborn-tx - Are you happy to resolve this one as the latest code has this change.

sergioprado:
@MMG-DG if you are confident you solved the problem raised by one reviewer, you can resolve the conversation yourself. No problem with that. If the reviewer finds out the issue was not really solved, then he/she can reopen it.

Also, I can see your PR has 10 commits. And you are implementing a single feature. If all changes are related to each other, there should be just one commit. In this case, please squash all commits into one.

Thanks!

MMG-DG:
> Also, I can see your PR has 10 commits. And you are implementing a single feature. If all changes are related to each other, there should be just one commit. In this case, please squash all commits into one.

Once everyone is happy with the final state of things I will "squash" the commits.
However, I believe there is one thing remaining to figure out and that is how to get raised errors from the piped commands.
For me the previous suggestion fails with errors.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +67,29 @@ tdx_enc_key_gen() {
...
+    if [ ${TDX_ENC_PRESERVE_DATA} -eq 1 ]; then
+        mkdir -p "${TDX_ENC_STORAGE_MOUNTPOINT}"
+        if mount ${TDX_ENC_STORAGE_LOCATION} "${TDX_ENC_STORAGE_MOUNTPOINT}"; then
+            tdx_enc_log "Backing up original content..."
+
+            TDX_ENC_PRESERVE_LOCATION=$(mktemp)
+            BACKUP_STORAGE_LIMIT="$(awk '/MemFree/ { printf "%lu", $2/30 }' /proc/meminfo)"
+            tdx_enc_log "Backup limit determined: ${BACKUP_STORAGE_LIMIT}"
+            
+            msgs="$(tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -c . | bzip2 -cz | dd bs=1024 count=${BACKUP_STORAGE_LIMIT} of=${TDX_ENC_PRESERVE_LOCATION})"
```
MMG-DG, line 81: Not sure what to do here to ensure the errors are raised and captured properly.
See: https://github.com/toradex/meta-toradex-security/pull/27#discussion_r1606716879

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -94,6 +123,25 @@ tdx_enc_partition_mount() {
...
 
+# restore data if available
+tdx_enc_restore_data() {
+    if [ ${TDX_ENC_PRESERVE_DATA} -eq 1 ]; then
+        tdx_enc_log "Checking backup file: ${TDX_ENC_PRESERVE_LOCATION}"
+
+        if [ -f ${TDX_ENC_PRESERVE_LOCATION} ]; then
+            tdx_enc_log "Restoring original content..."
+
+            msgs="$(bzip2 -cd ${TDX_ENC_PRESERVE_LOCATION} | tar -C ${TDX_ENC_STORAGE_MOUNTPOINT} -xf -)"
```
MMG-DG, line 134: Not sure what to do here to ensure the errors are raised and captured properly.
See: https://github.com/toradex/meta-toradex-security/pull/27#discussion_r1606716879

MMG-DG (commented):
Help needed to resolve requirements.

MMG-DG:
Just a quick note here, this code as-is has now been tested in my environments and seems to work as expected.

sergioprado:
> Once everyone is happy with the final state of things I will "squash" the commits.

I would prefer if you could squash them before another round of review.

I usually review a PR commit by commit, and this PR is kind of confusing,especially because there are commits that fix/change the code introduced by another commit in the same PR, like this one https://github.com/toradex/meta-toradex-security/pull/27/commits/9ac5a5c3e217147abf3f3d245d7267cad7d6f879.

Remember that you can always `git commit --amend` and `git push --force` to update the commit and push an updated version here.

For squashing, that should be straightforward via `git rebase -i`. Please review the final commit before submitting it.

Thanks!

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +67,30 @@ tdx_enc_key_gen() {
...
 
+# backup original data in partition (if not encrypted)
+tdx_enc_backup_data() {
+    if [ ${TDX_ENC_PRESERVE_DATA} -eq 1 ]; then
+        mkdir -p "${TDX_ENC_STORAGE_MOUNTPOINT}"
+        if mount ${TDX_ENC_STORAGE_LOCATION} "${TDX_ENC_STORAGE_MOUNTPOINT}"; then
+            tdx_enc_log "Backing up original content..."
+
+            TDX_ENC_PRESERVE_LOCATION=$(mktemp)
+            BACKUP_STORAGE_LIMIT="$(awk '/MemFree/ { printf "%u", $2/30 }' /proc/meminfo)"
```
sergioprado, line 78: Can we do this without `awk`? If not, you must add `awk` as a runtime dependency in the recipe.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
sergioprado: Not sure if @rborn-tx has any comments. I will have to think about this and do some tests. I will do this earlier next week.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -17,9 +17,15 @@ TDX_ENC_STORAGE_MOUNTPOINT="@@TDX_ENC_STORAGE_MOUNTPOINT@@"
 # dm-crypt device to be created
 TDX_ENC_DM_DEVICE="encdata"
 
+# Flag to enable preservation of data on partition before encryption
+TDX_ENC_PRESERVE_DATA=@@TDX_ENC_PRESERVE_DATA@@
+
+# storage location of data backup file (if needed)
+TDX_ENC_PRESERVE_LOCATION="/tmp/encdata.tar.bz2"
```
sergioprado, line 24: What about renaming this variable to `TDX_ENC_BACKUP_LOCATION` or `TDX_ENC_BACKUP_FILE`? I think it makes more clear to read the code.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -94,6 +124,26 @@ tdx_enc_partition_mount() {
...
 
+# restore data if available
+tdx_enc_restore_data() {
+    if [ ${TDX_ENC_PRESERVE_DATA} -eq 1 ]; then
+        tdx_enc_log "Checking backup file: ${TDX_ENC_PRESERVE_LOCATION}"
+
+        if [ -f ${TDX_ENC_PRESERVE_LOCATION} ]; then
+            tdx_enc_log "Restoring original content..."
+
+            msgs="$(bzip2 -cd ${TDX_ENC_PRESERVE_LOCATION} | tar -C ${TDX_ENC_STORAGE_MOUNTPOINT} -xf -)"
```
sergioprado, line 135: Is `bzip2` and `tar` part of BusyBox or they are provided by individual packages? If they are part of standalone packages, then you should add them to the list of runtime dependencies in the recipe

sergioprado:
@MMG-DG I added a few minor comments in the PR. I plan to do some tests on this earlier next week.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: awk is part of BusyBox. I was wondering if I should add a cfg file for that to ensure it gets included? Would be the same for grep too.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: yes they are. same as awk and grep for oe-core meta layers.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: Will update to use the `TDX_ENC_BACKUP_FILE` in my next squash.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: I only did it this way as it is much easier to read and less pipes to achieve the desired result. Other option was to use grep as in @sergioprado original comment:

> `# grep MemFree: /proc/meminfo | tr -s ' ' | cut -d ' ' -f 2`

However, that doesn't do the 30% calculation.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
rborn-tx: At this point I wanted to confirm: hasn't any of the options I mentioned along the way worked? As a recap, the options were:

1) To deal with the lack of pipefail in the shell:

```
msgs="$({ { tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -cj . || echo "ERROR" >&2; } | dd bs=1024 count=100 of=/tmp/saved.tar.bz2; } 2>&1)"
if [ "$?" -ne 0 ] || echo "${msgs}" | grep -q ERROR; then
    # Some error happened.
fi
```

2) To deal with the above plus the apparent unavailable `-j` option in `tar`:

```
msgs="$({ { tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -c . || echo "ERROR" >&2; } | { bzip2 -cz || echo "ERROR" >&2; } | dd bs=1024 count=100 of=/tmp/saved.tar.bz2; } 2>&1)"
if [ "$?" -ne 0 ] || echo "${msgs}" | grep -q ERROR; then
    # Some error happened.
fi
```

3) To deal with the above plus the apparent impossibility of capturing stderr into a shell variable:

```
{ { tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -c . || echo "ERROR" >&2; } | { bzip2 -cz || echo "ERROR" >&2; } | dd bs=1024 count=100 of=/tmp/saved.tar.bz2; } 2>/tmp/errors.txt
if [ "$?" -ne 0 ] || cat /tmp/errors.txt | grep -q ERROR; then
    # Some error happened.
fi
```

If none of them worked, then I ran out of ideas and would need to debug the actual script.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
rborn-tx: I believe we want the feature to work also on Torizon OS images so minimizing the dependencies is a good thing. Depending on coreutils and grep should be okay.

The below code works on the busybox `sh` of the reference image:

```
BACKUP_STORAGE_PCT=30    # Configurable percentage.
...
MEM_FREE=$(grep MemFree: /proc/meminfo | tr -s ' ' | cut -d ' ' -f 2)
BACKUP_STORAGE_LIMIT=$((MEM_FREE * BACKUP_STORAGE_PCT / 100))
```

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: @rborn-tx - Every time I have added the `.. || echo "ERROR" >&2` I get errors when the script runs.
I also had to split apart the bzip from the tar command (as it never seems to work), however, that is already in the latest code in my branch. I can only assume this is another BusyBox config that I have not yet found to enable the single tar command to work.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
MMG-DG: Latest test seems to work with these changes. Have updated my branch so this should now be included in latest PR changes.

rborn-tx:
@MMG-DG @sergioprado FYI, I've executed a test creating a service on a "standard" tdx reference image by defining a service with the exact same tdx-enc-handler.service file as in this PR and with the following script replacing tdx-enc.sh:

```
#!/bin/sh

# directory to mount the encrypted storage
TDX_ENC_STORAGE_MOUNTPOINT="/sbin/"

# log to standard output
tdx_enc_log() {
    echo "CAAM: $*"
}

# log error message and exit
tdx_enc_exit_error() {
    tdx_enc_log "ERROR: $*"
    exit 1
}

tdx_enc_main_start() {
    tdx_enc_log "Starting..."

    TDX_ENC_BACKUP_FILE=/tmp/backup
    BACKUP_STORAGE_LIMIT=2048          # 512 to cause the error
    tdx_enc_log "Backup limit: ${BACKUP_STORAGE_LIMIT}"

    # msgs="$(tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -cj . | dd bs=1024 count=${BACKUP_STORAGE_LIMIT} of=${TDX_ENC_BACKUP_FILE})" => Does not detect the error (as expected)
    msgs="$({ { tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -cj . || echo "ERROR" >&2; } | dd bs=1024 count=${BACKUP_STORAGE_LIMIT} of=${TDX_ENC_BACKUP_FILE}; } 2>&1)"
    if [ "$?" -ne 0 ] || echo "${msgs}" | grep -qi 'error\|invalid'; then
	tdx_enc_exit_error "Couldn't save original data."
    fi

    tdx_enc_log "Started successfully."
}

tdx_enc_main_stop() {
    tdx_enc_log "Finished"
}

tdx_enc_main() {
    case $1 in
        start)
            tdx_enc_main_start
            ;;
        stop)
            tdx_enc_main_stop
            ;;
        *)
            tdx_enc_exit_error "Invalid option! Please use 'start' or 'stop'."
            ;;
    esac

    tdx_enc_log "Success!"
}

tdx_enc_main "$1"
```

Rebooting the device with the backup limit set to 2048:

```
root@colibri-imx6-10866313:~# journalctl --no-pager -u tdx-enc-handler
Apr 28 17:42:28 colibri-imx6-10866313 systemd[1]: Starting Encryption handler for Toradex modules...
Apr 28 17:42:28 colibri-imx6-10866313 tdx-enc.sh[202]: CAAM: Starting...
Apr 28 17:42:28 colibri-imx6-10866313 tdx-enc.sh[202]: CAAM: Backup limit: 2048
Apr 28 17:42:34 colibri-imx6-10866313 tdx-enc.sh[202]: CAAM: Started successfully.
Apr 28 17:42:34 colibri-imx6-10866313 tdx-enc.sh[202]: CAAM: Success!
Apr 28 17:42:34 colibri-imx6-10866313 systemd[1]: Finished Encryption handler for Toradex modules.
```

Rebooting the device with the backup limit set to 512 (not enough to save the contents of /sbin/):

```
root@colibri-imx6-10866313:~# journalctl --no-pager -u tdx-enc-handler
Apr 28 17:42:28 colibri-imx6-10866313 systemd[1]: Starting Encryption handler for Toradex modules...
Apr 28 17:42:28 colibri-imx6-10866313 tdx-enc.sh[202]: CAAM: Starting...
Apr 28 17:42:28 colibri-imx6-10866313 tdx-enc.sh[202]: CAAM: Backup limit: 512
Apr 28 17:42:34 colibri-imx6-10866313 tdx-enc.sh[202]: CAAM: ERROR: Couldn't save original data.
Apr 28 17:42:34 colibri-imx6-10866313 systemd[1]: tdx-enc-handler.service: Main process exited, code=exited, status=1/FAILURE
Apr 28 17:42:34 colibri-imx6-10866313 systemd[1]: tdx-enc-handler.service: Failed with result 'exit-code'.
Apr 28 17:42:34 colibri-imx6-10866313 systemd[1]: Failed to start Encryption handler for Toradex modules.
```

That is, the results are exactly as expected.

This leads me to the question:

@MMG-DG What image are you testing? Are you using class tdxref-signed?

recipes-core/busybox/busybox/tdx-enc-handler-requirements.cfg
```diff
@@ -0,0 +1,19 @@
+#
+# Finding Utilities
+#
+CONFIG_GREP=y
+CONFIG_FEATURE_GREP_CONTEXT=y
```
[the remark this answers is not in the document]
MMG-DG: Oops... that was from something else I think. Will remove and see what happens.

recipes-core/busybox/busybox/tdx-enc-handler-requirements.cfg
```diff
@@ -0,0 +1,19 @@
...
+#
+# Archival Utilities
+#
+CONFIG_FEATURE_SEAMLESS_BZ2=y
+CONFIG_TAR=y
+CONFIG_FEATURE_TAR_LONG_OPTIONS=y
+CONFIG_FEATURE_TAR_CREATE=y
+CONFIG_FEATURE_TAR_AUTODETECT=y
+CONFIG_FEATURE_TAR_FROM=y
+CONFIG_FEATURE_TAR_GNU_EXTENSIONS=y
```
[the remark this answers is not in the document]
MMG-DG: Oops... that was from something else I think. Will remove and see what happens.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +70,31 @@ tdx_enc_key_gen() {
...
+        if mount ${TDX_ENC_STORAGE_LOCATION} "${TDX_ENC_STORAGE_MOUNTPOINT}"; then
+            tdx_enc_log "Backing up original content..."
+
+            TDX_ENC_BACKUP_FILE=$(mktemp)
+            MEM_FREE=$(grep MemFree: /proc/meminfo | tr -s ' ' | cut -d ' ' -f 2)
+            BACKUP_STORAGE_LIMIT=$((MEM_FREE * TDX_ENC_BACKUP_STORAGE_PCT / 100))
+            tdx_enc_log "Backup limit determined: ${BACKUP_STORAGE_LIMIT}"
+            
+            msgs="$(tar -C "${TDX_ENC_STORAGE_MOUNTPOINT}" -c . | bzip2 -cz | dd bs=1024 count=${BACKUP_STORAGE_LIMIT} of=${TDX_ENC_BACKUP_FILE})"
+            if [ "$?" -ne 0 ] || echo "${msgs}" | grep -q -i -E 'error|invalid'; then
```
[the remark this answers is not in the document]
MMG-DG: Will try it and see what happens.

MMG-DG:
> [ "$?" -ne 0 ]

I believe this is why it appears to work and get into the error branch. However, when I last tested that format the actual error that got returned was to do with the statement being malformed.

I will change this and try again to see if I can get the actual output and error messages.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc-caam.sh
```diff
@@ -61,6 +70,31 @@ tdx_enc_key_gen() {
     fi
 }
 
+# backup original data in partition (if not encrypted)
+tdx_enc_backup_data() {
+    if [ ${TDX_ENC_PRESERVE_DATA} -eq 1 ]; then
+        mkdir -p "${TDX_ENC_STORAGE_MOUNTPOINT}"
+        if mount ${TDX_ENC_STORAGE_LOCATION} "${TDX_ENC_STORAGE_MOUNTPOINT}"; then
+            tdx_enc_log "Backing up original content..."
```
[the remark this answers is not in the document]
MMG-DG: I think the `mkdir` line is still needed or the mount command would fail if the directory doesn't already exist (as is the case in my image).
I will apply this change and test this morning.

MMG-DG:
@rborn-tx - I am building the toradex-reference-minimal-image with tdx-signed. I have not yet had a chance to look at using the new tdxref-signed class yet.

MMG-DG:
> > @rborn-tx - I am building the toradex-reference-minimal-image with tdx-signed. I have not yet had a chance to look at using the new tdxref-signed class yet.
> 
> @MMG-DG I've repeated my test on a toradex-reference-minimal-image built with tdx-signed and again the results were as expected. Thus I can only think there is some other problem happening on your side.

Today I have re-run my tests and all seems to be working. I can only assume my original attempts had something slightly different that caused a syntax issue.

MMG-DG:
@rborn-tx @sergioprado 
I believe I have applied all suggestions again now and have been running the latest changes on a few images without issues.
I have not specifically tested the errors (as my images all seem to work without issues).

Hopefully this means it is almost there now.

rborn-tx (approved):
Latest version looks good to me, but I'll wait a bit to see if there's more feedback.

recipes-core/busybox/busybox_%.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+SRC_URI:append = "\
+    file://tdx-enc-handler-requirements.cfg \
```
sergioprado, line 4: This config fragment is only required if encryption is enabled, right?

If so, you can do something like this:

```
SRC_URI:append = "\
    ${@oe.utils.conditional('TDX_ENC_ENABLE', '1', 'file://tdx-enc-handler-requirements.cfg', '', d)} \
"
```

sergioprado:
@MMG-DG I have just one small recommendation.

Also, please add a Signed-off by in the commit message. You can do this by committing with `git commit -s`.

After this 2 changes, I think we are ready to merge this.

Thanks!

MMG-DG:
@sergioprado - Changes applied and I believe it is now "signed off".

sergioprado:
@MMG-DG everything looks good now. Thanks a lot for your effort during the review process and the contribution to the project!

PR #28, opened 2024-05-15
Introduce documentation about encryption

sergioprado:
While I was writing the documentation, i found out a few opportunities to improve the variables to configure this feature, so you might see a couple of patches renaming/changing variable names and values.

Also, I created a docs directory to organize better the documentation.

In case you want to see how the new markdown files are rendered, please check my fork:

https://github.com/sergioprado/meta-toradex-security/tree/enc-docs


docs/README-encryption.md
```diff
@@ -0,0 +1,57 @@
...
+
+Trusted Keys require the availability of a Trust Source for greater security, and CAAM is leveraged on iMX-based SoMs. TPM (Trusted Platform Module) and TEE (Trusted Execution Environment) are also two other possible trust sources for Trusted Keys, but currently not supported by this layer.
+
+## Block device encryption
+
+The Linux kernel [dm-crypt](https://docs.kernel.org/admin-guide/device-mapper/dm-crypt.html) subsystem is used to provide transparent disk encryption to users.
+
+`dm-crypt` is part of the device-mapper framework and works by creating a virtual block device that encrypts data as it is written to the disk and decrypts it as it is read.
+
+## Enabling and configuring encryption
```
jsrc27, line 41: Should we also mention in this section that `TDX_ENC_STORAGE_LOCATION` must be explicitly set to some valid value?

sergioprado: Makes sense! PR updated.

jsrc27 (approved):

docs/README-encryption.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+Here are the complete list of variables that can be used to customize the behavior of this feature:
+
+| Variable | Description | Default value |
+| :------- | :---------- | :------------ |
+| TDX_ENC_KEY_BACKEND | Backend used to manage the encryption key. Allowed values: `caam` or `cleartext`. If configured with `caam`, it will use Trusted Keys backed by the CAAM device (available on iMX-based SoMs). If configured with `cleartext`, the encryption key will be stored in clear text in the file system (use `cleartext` only for testing purposes!) | `caam` on iMX based SoMs, empty otherwise |
+| TDX_ENC_KEY_DIR | Directory to store the encryption key blob | `/var/local/private/.keys` |
+| TDX_ENC_KEY_FILE | File name of the encryption key blob | `key.blob` |
+| TDX_ENC_STORAGE_LOCATION | Partition to be encrypted (e.g. `/dev/sdb1`) | Empty |
```
rborn-tx, line 62: Maybe we should have a footnote related to this item warning the user that the name here must be the one assigned by the kernel, or if one wants to set a name that relies on udev rules then they must review the systemd dependencies of the service to ensure the name is available (since the service runs on an early stage where not necessarily udev has run/settled).

Does that make sense?

README.md
```diff
@@ -38,133 +38,23 @@ This layer supports the following security features:
...
-- Colibri iMX7D (1GB eMMC variant only)
-- Colibri iMX8X
-- Verdin AM62
-- Verdin iMX8MM
-- Verdin iMX8MP
+| Documentation | Description |
+| :------------ | :---------- |
+| [docs/README-secure-boot.md](docs/README-secure-boot.md) | General documentation about the secure boot feature |
+| [docs/README-secure-boot-imx.md](docs/README-secure-boot-imx.md) | Details on the secure boot implementation for iMX based SoMs |
+| [docs/README-secure-boot-k3.md](docs/README-secure-boot-k3.md) | Details on the secure boot implementation for K3 based SoMs (e.g. AM62) |
```
rborn-tx, line 51: Suggestion (as done in the README-secure-boot document):

- "iMX based" -> "NXP iMX-based"
- "K3 based" -> "TI K3-based"

rborn-tx (commented):
Nice refactoring of the docs!

I've added a couple suggestions for evaluation.

classes/tdx-encrypted.bbclass
```diff
@@ -9,21 +9,20 @@ TDX_ENC_ENABLE = "1"
...
-# Location in the filesystem to store the encryption key
-# Required only for the 'cleartext' backend
-TDX_ENC_KEY_FILE ?= "/run/.tdx-enc-key"
+# directory to store the encryption key blob
+TDX_ENC_KEY_DIR ?= "/var/local/private/.keys"
 
-# Directory to store CAAM keys and blobs (only for the CAAM backend)
-TDX_ENC_CAAM_KEYBLOB_DIR ?= "/var/local/private/caam/keys/"
+# encryption key blob file name
+TDX_ENC_KEY_FILE ?= "key.blob"
```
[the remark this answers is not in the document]
sergioprado: Makes sense! I changed to `tdx-enc-key.blob`.

docs/README-encryption.md
sergioprado: Oh yeah, I forgot about it. Thanks for the reminder. I added a footnote in the documentation.

README.md
sergioprado: Done!

sergioprado:
@rborn-tx I submitted another PR for review with your comments addressed. Thanks!

rborn-tx:
@sergioprado Thanks for the updates. Merging now.

PR #29, opened 2024-05-23
toradex-security: Add dynamic layers.

drewmoseley:
Now that this layer is being used in the common-torizon project, we are getting warnings about bbappend files with no matching bb file.  Using dynamic layers allows this to work in those configs without any warnings.

Signed-off-by: Drew Moseley <drew@moseleynet.net>


dynamic-layers/toradex-bsp-common/recipes-kernel/linux/device-tree-overlays-mainline_%.bbappend
```diff
@@ -0,0 +1 @@
+require ${@ 'recipes-kernel/linux/add-secboot-kargs-overlay.inc' if 'tdx-signed' in d.getVar('OVERRIDES').split(':') else ''}
```
sergioprado, line 1: @drewmoseley Why changing from `add-secboot-kargs-overlay.inc` to `recipes-kernel/linux/add-secboot-kargs-overlay.inc`? Usually `include` and `require` directives are able to find files if the included files are in the same directory as the file that it is including it.

Also, because you renamed the file and changed it in the same commit, Git thinks you are removing the old file and adding the new one. With this, we lose the history of the file. If this change is really required, would you please separate it in two commits, one to move the file and another to change it. This way, we can recover the history of the file with `git log --follow`.

Thanks!

drewmoseley: @sergioprado I had to change the include path because the files are not in the same directory as the bbappend.  In this case the include file stays in the main `recipes-kernel/linux/` directory and is included with a full path.  Since the include file is used by both device-tree-overlays and device-tree-overlays-mainline I think this makes sense.

I opted to both move the file and edit it in one step to avoid potential breakage of `git bisect` but if you are ok with that I can redo it as 2 separate commits.

sergioprado: > @sergioprado I had to change the include path because the files are not in the same directory as the bbappend. In this case the include file stays in the main recipes-kernel/linux/ directory and is included with a full path. Since the include file is used by both device-tree-overlays and device-tree-overlays-mainline I think this makes sense.

Hmm, that is right. You moved only the append file. Any reason for not moving the `add-secboot-kargs-overlay.inc` as well, so this change would not be required?

> I opted to both move the file and edit it in one step to avoid potential breakage of git bisect but if you are ok with that I can redo it as 2 separate commits.

Yeah, I understand that there is a trade-off. @rborn-tx do you have any thoughts on this?

drewmoseley: The include file is used by both `dynamic-layers/toradex-bsp-common/recipes-kernel/linux/device-tree-overlays-mainline_%.bbappend` and `dynamic-layers/toradex-nxp/recipes-kernel/linux/device-tree-overlays_%.bbappend` so I left it in the main `recipes-kernel/linux` folder so both could use it.

For this particular change, I can always do two commits; first with the change to the full path and second of the move.  That should preserve the "follow" history as well as the "git bisect"ability.  Let me know if you want me to modify it as such.

jsrc27: Just to add my 2 cents. If possible I prefer to keep the follow history of the files. Though this isn't a hard requirement from me or anything, just a preference.

sergioprado: Yeah, I also think splitting the PR in two commits to keep the history provides more benefits than costs.

@drewmoseley would you please split this PR in two commits and submit an updated version?

drewmoseley: @sergioprado,  this is done.

sergioprado:
@rborn-tx any comments before we merge this?

rborn-tx:
@drewmoseley I guess these changes are low risk, but I just wanted to ask: have you done a test build of Torizon OS with class `tdx-signed` after making the changes? And a test build of a reference image both with `tdx-signed` and `tdxref-signed`?

drewmoseley:
@rborn-tx no I have only built the common torizon stuff I was using. I assumed we would get test builds out of CI before this got completely merged. Is that assumption incorrect?

rborn-tx:
> @rborn-tx no I have only built the common torizon stuff I was using. I assumed we would get test builds out of CI before this got completely merged. Is that assumption incorrect?

@drewmoseley Unfortunately it is. Our CI only runs on stuff already merged. And today, the only configuration that is built and tested in CI is image `torizon-core-docker` with the class `tdx-signed`; there are not build or runtime tests for the BSP images.

It would be nice if you could at least build a BSP image for an NXP machine with both `tdx-signed` and `tdxref-signed` since those are not covered by CI. We could leave the build/runtime test to be done by CI in case of the Torizon OS image and possibly fix any issues afterwards.

Would you mind to do that before the merge?

rborn-tx:
> Of course. I'll do that in the morning. Is there anything beyond basic build and boot you think would be helpful?

For `tdx-signed` I'd check to see if the system is booting a signed FIT image.
For `tdxref-signed` I'd imagine checking if dm-verity is in use.

sergioprado:
@drewmoseley to test dm-verity, you might want to check if the DM target was created and if the rootfs mount point is pointing to it:

```
# ls -la /dev/mapper/rootfs
# cat /proc/mounts | grep " / "
```

Also, if you add the cryptsetup package to the image, you can check the DM verity status with the following command:

```
# veritysetup status /dev/dm-0
```

drewmoseley:
OK.  And how do I test if a signed FIT image is in use?  Will that just show in the UBoot logs that the signatures are being verified?

And to confirm, this will happen on both Torizon and Reference images right?

rborn-tx:
> OK. And how do I test if a signed FIT image is in use? Will that just show in the UBoot logs that the signatures are being verified?

Yes, the U-Boot logs will show information about the FIT image and you should see something as I highlighted below:

```
## Loading kernel from FIT Image at 50300000 ...
   Using 'conf-freescale_imx8mm-verdin-nonwifi-dev.dtb' configuration
   Verifying Hash Integrity ... sha256,rsa2048:dev+ OK     <=== (*)
   Trying 'kernel-1' kernel subimage
     Description:  Linux kernel
     Type:         Kernel Image
     Compression:  gzip compressed
```

The messages will be almost the same between torizon and the reference image. In torizon (only) there will be this line at the end of the logs because the hardening is enabled:

```
## Validation of bootargs succeeded.
```

drewmoseley:
OK.  After finally massaging all my build configs, rebasing my changes on the latest and getting successful builds, this all looks good.  I tested the apalis-imx8, apalis-imx6 and verdin-imx8mm machines.

With the torizon and tdxref builds, without tdx-signed/tdxref-signed, all build and boot as expected.

With tdx-signed/tdxref-signed, I verified:
* ahab_status or hab_status command at the UBoot prompt.
* Booting and verifying a signed FIT image
* dm-verity setup properly on tdxref images.

I'll remove RFC from the subject here since it seems this is about ready for merge.

sergioprado:
@drewmoseley thanks for testing! All looks good to me.

I will wait for @jsrc27 and @rborn-tx approval before merging this.

sergioprado:
@jsrc27 @rborn-tx a gentle reminder for you to review this one. I am planning to merge it tomorrow.

jsrc27 (approved):
I have nothing further to add.

rborn-tx (approved):
LGTM

PR #31, opened 2024-06-07
Add OP-TEE support for Verdin iMX8MP

sergioprado:
Commits in this PR:

- u-boot: change Verdin iMX8MP DRAM banks when building with OP-TEE
- linux: add config fragment for OP-TEE
- optee: add append files for OP-TEE recipes
- tdx-optee.bbclass: introduce class to enable OP-TEE support
- README.md: add documentation about OP-TEE

Build and tested on Verdin iMX8MP.


classes/tdx-optee.bbclass
```diff
@@ -0,0 +1,40 @@
...
+addhandler validate_optee_support
+validate_optee_support[eventmask] = "bb.event.SanityCheck"
+python validate_optee_support() {
+    supported_machines = ['verdin-imx8mp']
+    machine = e.data.getVar('MACHINE')
+    if machine not in supported_machines:
+        bb.fatal("OP-TEE is currently not supported on '%s' machine!" % machine)
+
+    if 'tdx-signed' in d.getVar('OVERRIDES').split(':'):
+        bb.warn("Secure boot might not work because the OP-TEE firmware is not being signed!")
```
jsrc27, line 36: Could you elaborate more on the possible consequences here?

When you say "might not work", is it not deterministic what would happen here?

jsrc27 (commented):
Looks great for the most part just left some questions I had.

classes/tdx-optee.bbclass
sergioprado: Hmmm. Good catch. Wrong wording indeed. I changed to "Secure boot will not work because the OP-TEE firmware is currently not being signed!". PR updated.

jsrc27: Thanks for the clarification. So then, if this will not work at all is `bb.warn` still appropriate here or would it make more sense to be `bb.fatal`?

sergioprado: Allright, so I thought about it. Although I am already working on a fix for this, indeed it makes more sense for now to just fail with an error if both secure boot and OP-TEE are enabled. PR updated.

rborn-tx (approved):
Looks pretty good to me!

rborn-tx:
@sergioprado Out of curiosity, is there any reason the same implementation wouldn't work on a Verdin-iMX8MM?

sergioprado:
> Out of curiosity, is there any reason the same implementation wouldn't work on a Verdin-iMX8MM?

@rborn-tx each SoC in [optee_os](https://github.com/OP-TEE/optee_os/) might require a specific configuration to work. See for example in the makefile below that the configuration for machines in `mx8mm-flavorlist` differ from the machines in `mx8mp-flavorlist`:

https://github.com/OP-TEE/optee_os/blob/master/core/arch/arm/plat-imx/conf.mk

But probably (and hopefully) using the `mx8mmevk` machine configuration will work for `verdin-imx8mm`. :-)

jsrc27 (approved):

PR #33, opened 2024-06-13
Introduce class to create data partition

sergioprado:
This class will enable the creation of a data partition in the eMMC for Easy Installer images.

This will make it possible to have a read-write partition for situations where the rootfs is read-only (e.g. when dm-verity is enabled).

A few additional variables can be used to configure this feature:

- `TDX_TEZI_DATA_PARTITION_TYPE`: data partition filesystem type (ext2, ext3, ext4, fat, ubifs)
- `TDX_TEZI_DATA_PARTITION_LABEL`: data partition label
- `TDX_TEZI_DATA_PARTITION_AUTOMOUNT`: configure whether the data partition should be automatically mounted at boot time or not
- `TDX_TEZI_DATA_PARTITION_MOUNTPOINT`: data partition mount point
- `TDX_TEZI_DATA_PARTITION_MOUNT_FLAGS`: data partition mount flags (to be added to fstab)

This depends on a feature added to `image_type_tezi.bbclass` in `meta-toradex-bsp-common`, already merged upstream:

https://git.toradex.com/cgit/meta-toradex-bsp-common.git/commit/?h=kirkstone-6.x.y&id=bbb39c255936a130c18d4eb9410b9558fe5a137b

Build and runtime tested on Verdin iMX8MP.

### DIFF
--- a/classes/tdx-tezi-data-partition.bbclass
+++ b/classes/tdx-tezi-data-partition.bbclass
@@ -1,0 +1,33 @@
+# override for conditional assignment
+DISTROOVERRIDES:append = ":tdx-tezi-data-partition"
+
+# data partition filesystem type
+# supported values: ext2, ext3, ext4, fat, ubifs
+# obs.: this is limited to what Easy Installer supports
+TDX_TEZI_DATA_PARTITION_TYPE ?= "ext4"
+
+# data partition label
+TDX_TEZI_DATA_PARTITION_LABEL ?= "DATA"
+
+# automount data partition at boot time
+TDX_TEZI_DATA_PARTITION_AUTOMOUNT ?= "1"
+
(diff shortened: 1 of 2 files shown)


jsrc27 (approved):
Looks alright to me. Good to see we got support for this from the Tezi side of things. I see the bbclass here only explicitly sets some of the `TEZI_DATA*` variables. I assume the defaults on the other variables are okay for most general purposes.

sergioprado:
> I assume the defaults on the other variables are okay for most general purposes.

Exactly! But of course it can be changed by the user if needed. I might mention this in the documentation, that I plan to write as soon as this is merged.

rborn-tx (approved):
LGTM as well

PR #35, opened 2024-06-25
Torizon rename

EdTheBearded:
Change the old TorizonCore name on our layer to the new Torizon OS.

Closes https://github.com/torizon/meta-toradex-torizon/issues/81

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This layer supports the following security features:
 
 The level of support of the above features is:
 
-- TorizonCore: tested and integrated.
+- Torizon: tested and integrated.
 - BSP Layers and Reference Images: not tested, integration effort is expected.
 
 The features are currently supported on the following SoMs:


sergioprado:
Would you please just remove "Closes #81" from the commit message, as this doesn't mean anything here in this layer?

EdTheBearded:
> Would you please just remove "Closes #81" from the commit message, as this doesn't mean anything here in this layer?

Done.

leograba:
Btw, I'm too late to this discussion, but for the future I'd like to let you know that using a full reference such as  `Closes torizon/meta-toradex-torizon#81` is possible.

That's the GitHub way to reference an issue on another organization and repo: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#issues-and-pull-requests

sergioprado:
> Btw, I'm too late to this discussion, but for the future I'd like to let you know that using a full reference such as Closes torizon/meta-toradex-torizon#81 is possible.

@leograba yes, if we can reference issues from other projects on GitHub (I guess that is the case now for meta-toradex-torizon), then it is certainly acceptable.

PR #38, opened 2024-07-05
Introduce support for TPMs to encrypt block devices

sergioprado:
Commits in this pull request:

- tdx-enc-handler: introduce support for TPM backend
- tdx-enc-handler: build kernel features as modules
- tdx-enc-handler: handle all backends in a single script


recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
```diff
@@ -53,6 +53,10 @@ tdx_enc_exit_error() {
 tdx_enc_prepare_generic() {
     tdx_enc_log "Preparing and checking system (generic)..."
 
+    if ! modprobe dm-crypt; then
```
rborn-tx, line 56: There's type on the commit message introducing this change: "inicialization"

rborn-tx (commented):
I couldn't go into much detail but overall it looks good to me. This is a nice addition to the layer!

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
```diff
@@ -37,39 +49,74 @@ tdx_enc_exit_error() {
...
 
-    ENC_KEY_FILE="${TDX_ENC_KEY_DIR}/${TDX_ENC_KEY_FILE}"
+    tdx_enc_log "Configuring key in kernel keyring (type=$TDX_ENC_KEY_KEYRING_TYPE keyname=$KEYNAME)..."
 
-    if [ ! -e "${ENC_KEY_FILE}" ]; then
-        tdx_enc_log "Encrypted key not found. Creating it..."
-        KEY="$(keyctl add trusted tdxenc 'new 32' @s)"
+    if [ ! -e "${TDX_ENC_KEY_FULLPATH}" ]; then
+        tdx_enc_log "Key blob not found. Creating it..."
+        KEYHANDLE="$(keyctl add "${TDX_ENC_KEY_KEYRING_TYPE}" "${KEYNAME}" "$(eval echo ${NEW_KEY_CMD})" @s)"
```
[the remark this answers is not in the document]
sergioprado: Yes, the `eval echo` is really required here.

If the variable would have just a string, then it would not be required, but sometimes the variable has a command that needs to be executed before being passed to the `keyctl` command.

For example, the `tdx_enc_key_gen_tpm()` function will pass "`load $(cat ${TDX_ENC_KEY_FULLPATH}`" to this variable, which needs to be expanded inside the tdx_enc_keyring_configure() function.

So the best solution I found for this is expanding the variable via `eval echo`. 

Obs.: I know shellchek complains about it.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
sergioprado: Good catch! Commit message updated.

jsrc27 (approved):
LGTM

PR #46, opened 2024-08-07
Import some fixes from the main branch

rborn-tx:
Import some fixes from main, namely:

- tdx-signed-harden.inc: update expected kargs for the apalis-imx8
- u-boot: update logic to detect bad usage of hardening


jsrc27 (approved):
LGTM

PR #48, opened 2024-08-11
Specialize HABv4 signing to i.MX6/7 in preparation for signing through U-Boot

rborn-tx:
Changes:

- Add a warning message on the `imx-boot` recipe to let users know signing will likely not work (relevant to the Toradex `verdin-imx8mm` and `verdin-imx8mp` modules).
- Since signing on i.MX8M is now supposed to be done through U-Boot (like with the i.MX6/7), here we make a small refactor by renaming scripts and templates to make them more specific to the i.MX6/7 in preparation for the additions that will in turn be specific to the i.MX8M.

By the way, the following commits in the PR are theoretically applicable also to `kirkstone`:

- u-boot-hab: make CSF-related source files specific to i.MX6/7
- u-boot-hab: make CSF creation script specific to i.MX6/7
- u-boot-hab: make signing function specific to i.MX6/7
- ~~imx-boot-hab: warn users about possibly non-working HABv4 signing~~
- imx-boot-hab: move duplicated code to common signing function

**Question**: do we want to cherry pick them to `kirkstone`?


recipes-bsp/u-boot/u-boot-hab.inc
```diff
@@ -23,9 +23,9 @@ UBOOT_BINARY_HAB = "${UBOOT_BINARY}"
 UBOOT_BINARY_HAB:mx6dl-generic-bsp = "u-boot-ivt.img"
 UBOOT_BINARY_HAB:mx6q-generic-bsp = "u-boot-ivt.img"
 
-# Sign U-Boot binary and/or SPL.
+# Sign U-Boot binary and/or SPL (this works both with i.MX6 and i.MX7).
 # This is supposed to run in the U-Boot build directory.
-sign_habv4() {
+imx6_sign_habv4() {
```
sergioprado, line 32: Since this also applies to iMX7, makes sense o rename it to something like `imx6_imx7_sign_habv4`?

sergioprado (commented):
I have just one comment, but looks good overall.

Related to the question about cherry-picking to kirkstone, I would say the answer is yes. Whatever we can cherry-pick to kirkstone, we should do it, as this might improve maintenance. The closer the branches are, the easier is to apply port changes to both branches.

Thanks!

recipes-bsp/u-boot/u-boot-hab.inc
```diff
@@ -5,27 +5,31 @@ DEPENDS += "util-linux-native"
 
 SRC_URI:append = "\
     file://u-boot-hab.cfg \
-    file://create_csf.sh \
     file://create_fuse_cmds.sh \
+"
+
+TDX_IMX6_IMX7_CSF_ARTIFACTS = "\
+    file://imx6_create_csf.sh \
```
jsrc27, line 12: Similar to Sergio's question. This script is for imx6 and imx7, correct?

recipes-bsp/u-boot/u-boot-hab.inc
rborn-tx: Absolutely! I'll do that.

recipes-bsp/u-boot/u-boot-hab.inc
rborn-tx: Correct. I'll make that change as well.

recipes-bsp/u-boot/u-boot-hab.inc
rborn-tx: Updated as per suggestion.

recipes-bsp/u-boot/u-boot-hab.inc
rborn-tx: Updated as per suggestion.

sergioprado:
LGTM

jsrc27 (approved):

PR #49, opened 2024-08-18
Update downstream hardening patches

rborn-tx:
Updating downstream U-Boot hardening patches due to the recent version bump at the BSP level from U-Boot v2022.04 to v2024.04. This should solve build errors on the "signed" images of the apalis-imx8 and colibri-imx8x.


jsrc27 (approved):
LGTM.

sergioprado (approved):
LGTM!

Issue #50, opened 2024-08-21
Remove hard dependencies on `meta-toradex-nxp` and `meta-toradex-ti`


tjallingt:
Hi, we'd like to use this layer in our custom yocto setup for the `verdin-am62`.
Currently to use this layer for the `am62` board we would need to install both the `meta-toradex-nxp` and `meta-toradex-ti` layers.
Since we build only for the ti board it seems silly to include a layer for the nxp based boards.
As far as I can tell all the board specific stuff is already located in `dynamic-layers` and only included conditionally based on the presence of either of these layers.

Would it be possible to move both of these to `LAYERRECOMMENDS` instead?

rborn-tx:
Closing this since the proposed fix was merged.

PR #51, opened 2024-08-21
Move toradex-nxp-layer and toradex-ti-layer to LAYERRECOMMENDS

tjallingt:
I'm not sure if this works as I'm imagining it should but this PR "downgrades" `toradex-nxp-layer` and `toradex-ti-layer` from being "hard" dependency (`meta-toradex-security` won't build without both of these present) to a recommendation (this layer will build without but recommends both, this way you can also build using only one of these). 

Issue: https://github.com/toradex/meta-toradex-security/issues/50

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,13 @@ revision: HEAD
 URI: git://git.yoctoproject.org/meta-security
 branch: kirkstone
 revision: HEAD
+```
 
+# Layer recommendations
+
+This layer recommends:
+
+```
 URI: git://git.toradex.com/meta-toradex-nxp.git
 branch: kirkstone
 revision: HEAD
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -10,9 +10,13 @@ BBFILE_PRIORITY_meta-toradex-security = "10"
 LAYERDEPENDS_meta-toradex-security = "\
     core \
     security \
+"
+
+LAYERRECOMMENDS_meta-toradex-security = "\
     toradex-nxp-layer \
     toradex-ti-layer \
 "
+
 LAYERSERIES_COMPAT_meta-toradex-security = "kirkstone"
 
 BBFILES_DYNAMIC += " \


sergioprado:
Makes sense to me. Out of curiosity, what is your use case? I would presume you are not using Toradex manifest files for the build, right?

I have a few suggestions on the pull request:

1. Change the title of the commit to `conf/layer.conf: move toradex-nxp-layer and toradex-ti-layer to LAYERRECOMMENDS`
2. Add a message in the commit describing why you are doing this change.
3. Adapt the README file. I would remove the layers from the `Layer dependencies` session, and add  them to a new session called `Layers recommendations`.

tjallingt:
Yea we have a yocto setup that uses the `meta-toradex-bsp-common` and `meta-toradex-ti` layers (but not any of the other layers that are installed by https://git.toradex.com/cgit/toradex-manifest.git/).

As for `meta-toradex-security` we are evaluating using secure boot for the verdin-am62 . At the moment that means we also need to add `meta-freescale-3rdparty`, `meta-freescale` and `meta-toradex-nxp` layers to our yocto setup even though we are only building for the am62.

.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
...
+  ],
+  "python.analysis.extraPaths": [
+    "${workspaceFolder}/sources/poky/bitbake/lib",
+    "${workspaceFolder}\\sources\\poky\\meta\\lib"
+  ],
+  "files.associations": {
+    "*.conf": "bitbake",
+    "*.inc": "bitbake"
+  }
+}
```
sergioprado, line 14: I guess you included this file in the commit by accident. Would you please remove it?

sergioprado:
I have just one smal comment about the inclusion of the vscode file. Also, would you please sign-off the commit by adding a tag with your name and email? Example:

```
Signed-off-by: Sergio Prado <sergio.prado@e-labworks.com>
```

.vscode/settings.json
tjallingt: ugh the bitbake vs code extension generated it and i didn't spot it sorry, ill remove it.

sergioprado (approved):
LGTM.

Will wait a bit to see if @jsrc27 or @rborn-tx have any feedback before merging this.

PR #52, opened 2024-08-22
tdx-enc-handler: add support for saving the encryption key in the disk partition

sergioprado:
On systems with secure boot enabled and an encrypted data partition, there is no place to store the encryption key blob, since the rootfs will be read-only (because of dm-verity) and one cannot use the data partition to store the key required to decrypt itself.

To solve this issue, let's make it possible to save the encryption key blob inside the encrypted partition, in a reserved block in raw mode.

When enabled (`TDX_ENC_KEY_LOCATION="partition"`), the last block of the partition to be encrypted will be reserved to store the encryption key blob, using the following format, where SIZE is a 3-bytes string representing the size of the key and KEY is the content of the encryption key blob:

```
'ENCKEY' + 'SIZE(3)' + KEY(N)
```

The default behavior is still to save the key as a file in the filesystem (`TDX_ENC_KEY_LOCATION="filesystem"`), so this feature will not break systems that are currently using encryption.

### DIFF
--- a/recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
+++ b/recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
@@ -5,6 +5,9 @@
 # backend used to manage the encryption key
 TDX_ENC_KEY_BACKEND="@@TDX_ENC_KEY_BACKEND@@"
 
+# encryption key location
+TDX_ENC_KEY_LOCATION="@@TDX_ENC_KEY_LOCATION@@"
+
 # directory to store CAAM encrypted key
 TDX_ENC_KEY_DIR="@@TDX_ENC_KEY_DIR@@"
 
@@ -13,6 +16,16 @@ TDX_ENC_KEY_FILE="@@TDX_ENC_KEY_FILE@@"
 
 # storage location to be encrypted (e.g. partition)
 TDX_ENC_STORAGE_LOCATION="@@TDX_ENC_STORAGE_LOCATION@@"
+
+# Number of blocks to reserve from the partition to be encrypted
+# Useful in case one needs a storage location to save data in raw
+# mode, outside the dm-drypt partition
+TDX_ENC_STORAGE_RESERVE="@@TDX_ENC_STORAGE_RESERVE@@"
+
+# number of blocks in the storage to be encrypted by dm-crypt
+# depends on the size of the partition and the number of blocks
+# reserved (see TDX_ENC_STORAGE_RESERVE), initialized at runtime
+TDX_ENC_STORAGE_NUM_BLOCKS=""
 
 # directory to mount the encrypted storage
 TDX_ENC_STORAGE_MOUNTPOINT="@@TDX_ENC_STORAGE_MOUNTPOINT@@"
@@ -60,6 +73,20 @@ tdx_enc_prepare_generic() {
     if ! dmsetup targets | grep crypt -q; then
         tdx_enc_exit_error "No support for dm-crypt target!"
     fi
+
+    TDX_ENC_STORAGE_NUM_BLOCKS=$(blockdev --getsz ${TDX_ENC_STORAGE_LOCATION})
+    if [ "${TDX_ENC_KEY_LOCATION}" = "partition" ] && [ "${TDX_ENC_STORAGE_RESERVE}" = "0" ]; then
+        # we need at least one reserved block to store the encryption key blob
+        TDX_ENC_STORAGE_RESERVE="1"
+    fi
+    TDX_ENC_STORAGE_NUM_BLOCKS=$((TDX_ENC_STORAGE_NUM_BLOCKS - TDX_ENC_STORAGE_RESERVE))
+
+    tdx_enc_log "Blocks to be encrypted: $TDX_ENC_STORAGE_NUM_BLOCKS..."
+    tdx_enc_log "Reserved blocks: $TDX_ENC_STORAGE_RESERVE..."
+
+    if [ "${TDX_ENC_KEY_LOCATION}" = "partition" ]; then
+        tdx_enc_key_recover_from_partition
+    fi
 }
 
 # CLEARTEXT: prepare system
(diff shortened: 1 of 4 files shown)


sergioprado:
I always wanted to be able to enable all security features at the same time (secure boot, encryption, OP-TEE). With this feature, we can. :-)

jsrc27 (approved):
LGTM.

My only suggestion and it's not a major one is. It would be nice if there was someway to notify the user about best practices. For example with this, we now recommend users to set the key location to partition if it's a read-only filesystem (i.e dm-verity). 

Though the user would only know about this if they had prior knowledge or read the relevant README which doesn't always happen. Not sure if there is a easier way to communicate these things.

sergioprado:
@jsrc27 thanks for reviewing!

> My only suggestion and it's not a major one is. It would be nice if there was someway to notify the user about best practices. For example with this, we now recommend users to set the key location to partition if it's a read-only filesystem (i.e dm-verity).

Yeah, that is a good question. I mention this when documenting the variable, but I am not sure if we should go beyond that. I think the layer should document the available features and provide some information about its interface and how to use it. But I don't think we should write tutorials here. Possibly the best place to document best practices, real use-cases, etc, would be on the developer website.

rborn-tx:
> @rborn-tx do you plan to review this one?

@sergioprado Sorry for the delay. Yes, I intend to look at it today.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
```diff
@@ -97,6 +116,59 @@ tdx_enc_prepare_tpm() {
     fi
 }
 
+tdx_enc_key_recover_from_partition() {
```
rborn-tx, line 119: Regarding `tdx_enc_key_recover_from_partition()` and `tdx_enc_key_save_to_partition()`:

Though I'm not strongly against the approach here where the data is stored in binary format, I'm thinking about reuse of the partition for other purposes (by us or by others) and for that I'd argue that a text format would be more flexible (because we could have multiple variable size fields in any order).

A good example of the text approach would be the U-Boot environment which is kept as a series of null-terminated strings whether stored on the filesystem or on a raw device.

If you were to consider following that approach I'd suggest storing the data just like a text file with lines terminated by a newline character and the whole file being terminated by a EOT (`^D`) character. We could read the expected maximum number of blocks for the partition and then trim the data at the first occurrence of the EOT. Each line would be in the form:

```
key=value\n
```

and `value` could have anything except newlines but in general should be readable text. For example the line with the encrypted key could look like this (with the value as a hex string):

```
encpartkey=ac6722fece9d207ef8b4d3949ff2a28f0020d95624c81624e4782f405ad98072\n
```

If for whatever reason we want some kind of validation for the presence of the data in the partition we could add some fake entry at the beginning such as:

```
_MAGIC=123456\n
...entries...
```

What do you think?

rborn-tx (commented):
@sergioprado This is a nice addition to the layer in terms of feature.

In terms implementation I've added a proposal for a different approach which I'd like to discuss before merging this.

Thanks!

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
```diff
@@ -60,6 +68,17 @@ tdx_enc_prepare_generic() {
     if ! dmsetup targets | grep crypt -q; then
         tdx_enc_exit_error "No support for dm-crypt target!"
     fi
+
+    TDX_ENC_STORAGE_NUM_BLOCKS=$(blockdev --getsz ${TDX_ENC_STORAGE_LOCATION})
+    if [ "${TDX_ENC_KEY_LOCATION}" = "partition" ]; then
+        # last block is used to store the encrypted key blob
+        TDX_ENC_STORAGE_NUM_BLOCKS=$((TDX_ENC_STORAGE_NUM_BLOCKS - 1))
```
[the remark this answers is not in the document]
sergioprado: Yeah, makes sense to make the implementation more future-proof. I will create a variable to make the number of reserved blocks configurable.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
sergioprado: Excelent idea! I haven't though about it. Will do in V2.

sergioprado:
@jsrc27 @rborn-tx V2 is ready for review, In this version:

- Added a variable called `TDX_ENC_STORAGE_RESERVE` to configure the number of blocks to reserve from the partition to be encrypted.
- Changed how the information about the encryption key blob is stored in the partition to a list of null-terminated strings
in the format `key=value`.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
```diff
@@ -97,6 +124,63 @@ tdx_enc_prepare_tpm() {
...
+
+    while read -r line; do
+        key=$(echo "$line" | cut -d'=' -f1)
+        val=$(echo "$line" | cut -d'=' -f2)
+        case "$key" in
+            "keyname") keyname="${val}" ;;
+            "keydata") keydata="${val}" ;;
+            "keycsum") keycsum="${val}" ;;
+        esac
+    done < ${STORAGE_KEY_BLOCK_DATA}
```
rborn-tx, line 146: Hey @sergioprado,

I did a couple tests and found that this loop will happily process the data after the EOT:

Running this:

```
{
    printf "key1=val1\n"
    printf "key2=\tstring with tabs and NUL \0\0\0\0 characters !\n"
    printf "key3=val3\n"
    printf "\04"
    printf "badkey1=data after EOT"
    printf "\0\0\0\0\0\0\0\0\0\0\0\0\n"
    printf "badkey2=data after EOT\n"
} > tt.tmp

while read -r line; do
    echo "line: $line"
done < tt.tmp
```

I got the output:

```
line: key1=val1
line: key2=	string with tabs and NUL  characters !
line: key3=val3
line: ^Dbadkey1=data after EOT
line: badkey2=data after EOT
```

A simple solution is (tested on bash, dash and busybox sh):

```
EOT=$(printf '\004')
while read -r line; do
    echo "$line" | grep -q "$EOT" && break
    echo "line: $line"
done < tt.tmp
```

where the output is as expected:

```
line: key1=val1
line: key2=	string with tabs and NUL  characters !
line: key3=val3
```

rborn-tx (changes requested):
@sergioprado Thanks for the quick update. Please see my feedback for V2; other than that, it looks good to me.

sergioprado:
@rborn-tx @jsrc27 V3 is ready for review. Changes in V3:

- Improved handling of formatted data in the partition, as suggested by Rogerio.

rborn-tx (approved):
Thanks for the updates. Approving it now!

PR #53, opened 2024-08-26
Add support for HAB and kernel FIT image signing when BINMAN is in use (i.MX8M)

rborn-tx:
As the title says, this is adding support for HAB and FIT signing when BINMAN is used for the generation of the bootloader container; this is a new, more powerful approach upstream is following for producing the container on the i.MX8M and it'll likely happen with newer SoCs. Before, the container was produced by the imx-boot recipe but now it's done by U-Boot itself which has BINMAN inside its code base.

This introduces the following commits:

1. u-boot-fit-signature: support FIT image signing alongside binman
2. u-boot-hab.cfg: explicitly set CONFIG_SPL_LOAD_FIT_ADDRESS
3. u-boot-hab.cfg: no longer disable CONFIG_SPL_USB_GADGET (i.MX8M)
4. u-boot-hab: introduce signing for flash.bin produced by binman

Please refer to the individual commit messages for details.

IMPORTANT: I've started a conversation with the BSP team to include the setting `CONFIG_SPL_LOAD_FIT_ADDRESS` (commit (2)) on a BSP layer but for now I'm leaving it here until we reach a conclusion.


rborn-tx:
Btw, I think commit (3) would solve issue #4.

recipes-bsp/u-boot/files/mx8m-generic-bsp/u-boot-hab.cfg
```diff
@@ -2,3 +2,4 @@ CONFIG_IMX_HAB=y
 CONFIG_SPL_DRIVERS_MISC=y
 CONFIG_SPL_CRYPTO=y
 CONFIG_ARCH_MISC_INIT=y
+CONFIG_SPL_LOAD_FIT_ADDRESS=0x44000000
```
jsrc27, line 5: The commit message for this says this is a "short term solution". Out of curiosity do you already have in mind a "long-term" solution for this?

recipes-bsp/u-boot/files/imx8m_sign.sh
```diff
@@ -0,0 +1,307 @@
+#!/bin/bash
```
jsrc27, line 1: This script now replaces the script at `dynamic-layers/freescale/recipes-bsp/imx-mkimage/files/mx8m_create_csf.sh`, right?

Should the other script be deleted now or is there a reason to keep it still?

recipes-bsp/u-boot/files/mx8m-generic-bsp/u-boot-hab.cfg
rborn-tx: Yes, depending on the result of the conversation with the BSP team (mentioned in the description) we should hopefully be able to move this setting to the defconfigs of verdins (or to a BSP layer). But the solution is the same, what changes is only where we'd apply it. Applying it on this security layer is arguably not ideal because the layer shouldn't be concerned with the location of artifacts unrelated to secboot on the device memory space.

recipes-bsp/u-boot/files/imx8m_sign.sh
rborn-tx: Yes, in the end it replaces that script.

Today, as done on a recent PR, we have a deprecation warning saying that the old signing method (based on the output of imx-boot) shouldn't be used and will likely not work. In fact, all the signing is working even with the old method, however the resulting binary does not work (it hangs) simply because the SPL grows to large at least with the default defconfig; though someone eager to make the old method work could try to remove SPL features to downsize it enough to make it boot appropriately. I imagine not many people would be willing to go through that hassle now that we have the new method working.

Considering the above, I think it would be perfectly fine to get rid of the old method even as part of this PR, but I wanted to hear opinions...

sergioprado: I think we can drop it as part of this PR. The old implementation will always be there in the kirkstone branch anyways, and we certainly don't want to keep two different implementations here.

sergioprado (commented):
@rborn-tx great work! 

I have just a couple of comments on top of @jsrc27 comments.

But in general, the implementation looks good to me, though I must confess I overlooked some of your bash/sed magic. :-)

recipes-bsp/u-boot/files/mx8m-generic-bsp/u-boot-hab.cfg
rborn-tx: I started the conversation and today I noticed they created a ticket to handle this on their side and it seems this will be given priority. The change will most likely go straight into the upstream U-Boot defconfig for the verdin-imx8m/p. I imagine it will take a couple of weeks until all work is done. I'm following up their ticket so that when it's done I'd make the change to the present layer as well.

recipes-bsp/u-boot/files/imx8m_sign.sh
rborn-tx: Thanks guys for the feedback. I'll make this change then.

rborn-tx: Done on v2.

jsrc27 (approved):

recipes-bsp/u-boot/files/mx8m-generic-bsp/u-boot-hab.cfg
sergioprado: Sounds good, thanks!

sergioprado:
@rborn-tx @jsrc27 I think we are ready now to fork `scarthgap` from `main`. What do you think?

On the secure boot side, only AHAB support is missing, right?

rborn-tx:
> @rborn-tx @jsrc27 I think we are ready now to fork `scarthgap` from `main`. What do you think?
> 
> On the secure boot side, only AHAB support is missing, right?

@sergioprado Yeah, it seems like a good time to fork it.

Regarding secure boot, in theory AHAB should be working since the signing didn't change (it still uses imx-boot) and my fix for the container generation is already available on `meta-freescale`.

sergioprado:
> Regarding secure boot, in theory AHAB should be working since the signing didn't change (it still uses imx-boot) and my fix for the container generation is already available on meta-freescale. Also, the U-Boot hardening patches have been already updated to the version being now used on the apalis-imx8 and colibri-imx8 (the two ones which still use downstream U-Boot (and AHAB)).

Nice!

> @sergioprado Yeah, it seems like a good time to fork it.

Alright. Done!

https://github.com/toradex/meta-toradex-security/tree/scarthgap-7.x.y

PR #55, opened 2024-08-28
tdx-enc-handler: add support for saving the encryption key in the disk partition

sergioprado:
On systems with secure boot enabled and an encrypted data partition, there is no place to store the encryption key blob, since the rootfs will be read-only (because of dm-verity) and one cannot use the data partition to store the key required to decrypt itself.

To solve this issue, let's make it possible to save the encryption key blob inside the disk partition to be encrypted, but in a reserved block outside the dm-crypt partition.

This feature can be enabled by setting the TDX_ENC_KEY_LOCATION variable to 'partition'.

When enabled, at least one block from the partition is reserved to store the encryption key blob. In case one wants more blocks to be reserved for other purposes, the number of blocks to be reserved can be configured via the TDX_ENC_STORAGE_RESERVE variable.

The key is stored in the partition as a list of null-terminated strings in the format key=value:

keyname=tdxenc
keydata=e2204090ffee9159ebaff4e322590c4a4f44f203f84c3411b73a6788c6ae1... keycsum=29d6ef285d4ea17b0ccf1d94c9fe3957058b2b7733459d33c3ed12f9d031408f

The default behavior is still to save the key as a file in the filesystem (TDX_ENC_KEY_LOCATION set to 'filesystem'), so this feature will not break systems that are currently using encryption.

### DIFF
--- a/classes/tdx-encrypted.bbclass
+++ b/classes/tdx-encrypted.bbclass
@@ -10,7 +10,15 @@ TDX_ENC_ENABLE = "1"
 TDX_ENC_KEY_BACKEND ?= ""
 TDX_ENC_KEY_BACKEND:imx-generic-bsp ?= "caam"
 
+# Encryption key blob location
+# This variable defines where the encrypted key will be stored
+# Available options:
+#    filesystem -> encrypted key blob is stored as a file in the filesystem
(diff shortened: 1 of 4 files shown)


sergioprado:
This PR is just a cherry-pick from kirkstone into scarthgap.

jsrc27 (approved):
LGTM

PR #56, opened 2024-08-30
u-boot-fit-signature-nxp: fix build error due to renamed fuction

rborn-tx:
With commit "u-boot-hab: make signing function specific to i.MX6/7" we renamed the signing function but missed the rename in the FIT signing module. This commit should fix this situation.


sergioprado:
LGTM

PR #57, opened 2024-09-03
u-boot: display warnings on disallowed commands (kirkstone)

### LINKED ISSUE #23: Display warnings on bootloader commands that wouldn't be allowed on closed state
The whitelisting feature implemented as part of our U-Boot hardening features works like this:

- If the device is **open** and the command passes the whitelist for the **open** state accept/execute the command without messages; if the command does not pass the said whitelist show a warning and do not execute the command.
- If the device is **closed** and the command passes the whitelist for the **closed** state accept/execute the command without messages; if the command does not pass the said whitelist show a warning and do not execute the command.

To detect situations where a new command is added to the boot script that would prevent boot after the device is closed, it would be useful to show a warning in the following situation:

- If the device is **open** and the command does not pass the whitelist for the **closed** state.

We might consider showing the message only on the first N occurrences to avoid bloating the logs.

Having this improvement should avoid the situation that was reported in issue #20.

rborn-tx:
With this implementation that is part of the hardening, U-Boot will show warning messages informing the users when a command that is accepted in open state would not be allowed when the device is closed. This is useful to let users know that closing the device would likely leave it in an unbootable state if any of the commands is part of their boot script.

Resolves #23

As an example of how this helps, running with the present implementation on a BSP image produces an output like this:

```
## WARNING: Command execution WOULD BE DENIED in closed state (blocked by category) for `part uuid mmc 0:2...`.
```

which shows that the boot script of a BSP image is running the "part uuid" command which would not be allowed in closed state; solving this will be the topic of an upcoming implementation.

IMPORTANT:

- Since the hardening is implemented as a series of patches, it may be easier to review/understand the changes in this dummy pull request (where the relevant changes are in the last commit): https://github.com/rborn-tx/u-boot/pull/1

### DIFF
--- a/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch
+++ b/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch
@@ -1,4 +1,4 @@
-From 90474be140b1898670bd6e23e27cca6be4deb57c Mon Sep 17 00:00:00 2001
+From 39ebb14f545ffaa15661d194f028213002a8a497 Mon Sep 17 00:00:00 2001
 From: Rogerio Guerra Borin <rogerio.borin@toradex.com>
 Date: Mon, 28 Aug 2023 13:28:50 -0300
 Subject: [PATCH 1/8] toradex: common: add command whitelisting modules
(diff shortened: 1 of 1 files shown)


sergioprado:
LGTM.

sergioprado (approved):

jsrc27 (approved):

PR #59, opened 2024-09-03
u-boot: display warnings on disallowed commands

rborn-tx:
Same as https://github.com/toradex/meta-toradex-security/pull/57 but for the main branch.

### DIFF
--- a/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch
+++ b/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch
@@ -1,4 +1,4 @@
-From 4edc9d9ddc7ea04059de0640d33cdc135562ce0a Mon Sep 17 00:00:00 2001
+From 31e1ab9d8cb3aa57eaa821607ef3d344ba572458 Mon Sep 17 00:00:00 2001
 From: Rogerio Guerra Borin <rogerio.borin@toradex.com>
 Date: Mon, 28 Aug 2023 13:28:50 -0300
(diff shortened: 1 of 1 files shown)


sergioprado:
@rborn-tx we haven't discussed this yet. But I think we should not maintain the main branch (there are no users of the main branch, no automated builds, etc). I think it is too much overhead with little benefits for now.

We can maintain just kirkstone and scarthgap, and leave the main branch as is. As soon as we decide to work on a new OE version, then we can cherry-pick from scarthgap to main and fork a new version, just as we have done with scarthgap.

Does it make sense? Or you see any advantages on maintaining the main branch as well?

rborn-tx:
> @rborn-tx we haven't discussed this yet. But I think we should not maintain the main branch (there are no users of the main branch, no automated builds, etc). I think it is too much overhead with little benefits for now.
> 
> We can maintain just kirkstone and scarthgap, and leave the main branch as is. As soon as we decide to work on a new OE version, then we can cherry-pick from scarthgap to main and fork a new version, just as we have done with scarthgap.
> 
> Does it make sense? Or you see any advantages on maintaining the main branch as well?

@sergioprado That seems perfectly sensible to me. We can even drop this PR and sync `main` in batches when needed.

sergioprado:
> @sergioprado That seems perfectly sensible to me. We can even drop this PR and sync main in batches when needed.

Perfect. I will close this PR then.

> P.S.: That reminds me that we should switch our manifests to using the scarthgap-7.x.y branch of meta-toradex-security rather than master; I'll go ahead and o that.

Yeah, that is the perfect time to do this. :-)

PR #60, opened 2024-09-11
Revert "u-boot-hab.cfg: explicitly set CONFIG_SPL_LOAD_FIT_ADDRESS"

rborn-tx:
This reverts commit `32121191f9627879e644826b74f44133b95d465b` now that the same setting is already applied by meta-toradex-bsp-common since commit `f09a579cd52008cb73d471631a8d5d7306763712`.


jsrc27 (approved):
LGTM

PR #61, opened 2024-09-16
conf/layer.conf: make layer compatible with styhead

EdTheBearded:
Adding compatibility to fix our master build.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_meta-toradex-security = "\
     toradex-ti-layer \
 "
 
-LAYERSERIES_COMPAT_meta-toradex-security = "scarthgap"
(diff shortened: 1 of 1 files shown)


conf/layer.conf
```diff
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_meta-toradex-security = "\
     toradex-ti-layer \
 "
 
-LAYERSERIES_COMPAT_meta-toradex-security = "scarthgap"
+LAYERSERIES_COMPAT_meta-toradex-security = "scarthgap styhead"
```
sergioprado, line 20: @EdTheBearded would you please remove `scarthgap` and leave only `styhead`?

sergioprado (changes requested):

conf/layer.conf
EdTheBearded: Sure, already updated the commit.

PR #67, opened 2024-09-26
u-boot: include OP-TEE firmware in the boot container image

sergioprado:
Toradex BSP now uses imx-boot-container class and binman to create the boot container for iMX8M, and thus changes are required in the U-Boot device tree so binman can include the OP-TEE firmware in the boot container image.

With this change, OP-TEE works as expected in the scarthgap branch.

Tested on Verdin iMX8MP.

### DIFF
--- a/recipes-bsp/u-boot/files/0001-imx8mp-binman-add-tee-node.patch
+++ b/recipes-bsp/u-boot/files/0001-imx8mp-binman-add-tee-node.patch
@@ -1,0 +1,50 @@
+From 0e44c1769d9c7be3b5df154a287424aa3e70013a Mon Sep 17 00:00:00 2001
+From: Sergio Prado <sergio.prado@e-labworks.com>
+Date: Wed, 25 Sep 2024 22:33:08 -0300
+Subject: [PATCH] imx8mp: binman: add tee node
+
+Upstream-Status: Pending
+
+Signed-off-by: Sergio Prado <sergio.prado@e-labworks.com>
+---
+ arch/arm/dts/imx8mp-u-boot.dtsi | 17 ++++++++++++++++-
+ 1 file changed, 16 insertions(+), 1 deletion(-)
+
+diff --git a/arch/arm/dts/imx8mp-u-boot.dtsi b/arch/arm/dts/imx8mp-u-boot.dtsi
+index c4c1a1771026..999e06ff63da 100644
+--- a/arch/arm/dts/imx8mp-u-boot.dtsi
++++ b/arch/arm/dts/imx8mp-u-boot.dtsi
+@@ -167,6 +167,19 @@
+ 				};
+ #endif
+ 
++				tee {
++					description = "TEE firmware";
++					type = "firmware";
++					arch = "arm64";
++					compression = "none";
++					load = <0x56000000>;
++					entry = <0x56000000>;
++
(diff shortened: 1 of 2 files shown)


rborn-tx:
@sergioprado Some questions:

- Shouldn't this be upstreamed?
- Shouldn't this be done on imx8mm as well (since both imx8mm and imx8mp are the ones using binman)?

rborn-tx:
Out of curiosity I decided to check how the tee node is normally named on existing U-Boot dtsi's (see [tee-occurrences.txt](https://github.com/user-attachments/files/17151855/tee-occurrences.txt) for raw data):

```
|-------------+------------------|
| Node name   | Num. occurrences |
|-------------+------------------|
| sandbox_tee |                2 |
| op-tee      |                1 |
| tee         |                8 |
| optee       |               58 |
|-------------+------------------|
```

Not that we need to follow the most common name, but just something to consider...

sergioprado:
> Some questions:

@rborn-tx two good questions... :-) 

> Shouldn't this be upstreamed?

I thought about it, but this change would require OP-TEE to be enabled in the BSP, or binman would fail because `tee.bin` would not be found. Also, upstreaming this would mean the default BSP would boot OP-TEE, and that is possibly not desired on most use cases, and probably the reason for the BSP team to deny it. One possible solution would be to use an #ifdef in the node to conditionally add it depending on a config option. We might want to try to upstream that. Anyway, I would recommend merging this first, because I think upstreaming might take a while due to the challenges mentioned. Makes sense?

> Shouldn't this be done on imx8mm as well (since both imx8mm and imx8mp are the ones using binman)?

Possibly yes. But since this layer only supports OP-TEE on iMX8MP, I focused on that. Keeping additional changes without any benefits would not make much sense.

sergioprado:
> Not that we need to follow the most common name, but just something to consider...

Actually, there is no standard name for this node. There is indeed a node called `optee` used a lot (58x in your statistics), but this node is for something else. This node is to enable OP-TEE support in U-Boot itself, so U-Boot can talk and send requests to OP-TEE. This node is something like this:

```
optee {
    compatible = "linaro,optee-tz";
    method = "smc";
};
```

The node I created is for instructing `binman` to include `tee.bin` in the FIT image of the boot container, and there is no default name for that.

jsrc27 (approved):
No further comments from my side.

rborn-tx:
> One possible solution would be to use an #ifdef in the node to conditionally add it depending on a config option. We might want to try to upstream that. Anyway, I would recommend merging this first, because I think upstreaming might take a while due to the challenges mentioned. Makes sense?

@sergioprado Yes the idea would be to protect the inclusion of the `tee` node with an `#ifdef` (to make this upstreamable). But for sure we don't need to wait for upstreaming before merging this here. I'd suggest only adding an `Upstream-Status: Pending` to the patch.

sergioprado:
> Yes the idea would be to protect the inclusion of the tee node with an #ifdef (to make this upstreamable). But for sure we don't need to wait for upstreaming before merging this here. I'd suggest only adding an Upstream-Status: Pending to the patch.

@rborn-tx Done!

In parallel, I am starting a discussion with the BSP team on upstreaming this. Thanks for reviewing!

rborn-tx (approved):

Issue #64, opened 2024-09-18
Secure boot failure if silent=1 specified in CONFIG_EXTRA_ENV_SETTINGS


MMG-DG:
Using an iMX8MP with kirkstone yocto build (based on tdx-minimal-reference-image)...

Having added `silent=1 loglevel=0 udev.log_level=0` settings to the u-boot environment, Should any critical errors be written to the console.

I expect the following line should be changed to print error and not just normal `printf` message to console?
[0008-toradex-integrate-bootargs-protection-downstream.patch](https://github.com/toradex/meta-toradex-security/blob/c05a06c4283beb2ad36e08d22d97eee3d52778f3/recipes-bsp/u-boot/files/0008-toradex-integrate-bootargs-protection-downstream.patch#L67)

I assume there is something that can allow this to still be printed as a critical error message?
At least I believe critical errors should still be surfaced? (I have seen Linux kernel errors when they are raised, but nothing from u-boot; maybe the `silent` setting can be removed if the `loglevel` one(s) will silence all output except for errors?).

MMG-DG:
Just an update from my local testing....

If I remove the `silent=1` from the u-boot commands then it would appear I still get full output (which shows a lot of memory addresses and information to a potential attacker). 
This means that for my case I will need to continue with setting this value and just hope that nothing else goes wrong when setting the required fuses.

On a positive note, it does appear that the issue I was seeing in my previous build was likely not related to bootargs and may have been an error in one of my systemd services causing a boot loop due to watchdog timeouts.

MMG-DG:
Addition to above from local testing...

In my latest testing, it would seem that if I add the `silent=1` to the `CONFIG_EXTRA_ENV_SETTINGS` in `include/configs/verdin-imx8mp.h`, that once I set the secure boot fuse, my device now ends up in an infinite boot loop with no output to determine why this is happening.
This is the only difference I have between a SoM that works (but shows a lot of information that shows memory addresses that could be used to attack the system) and a SoM the gets into a boot loop (which does not show any details.

This is my current patch being applied that doesn't work.
```
From 7ce11a7ac92bf4cae930f761ae0645f9dda6a1ff Mon Sep 17 00:00:00 2001
From: John Doe <johndoe@example.com>
Date: Sat, 24 Aug 2024 16:14:49 +0100
Subject: [PATCH 12/12] Silent uboot args

---
 include/configs/verdin-imx8mp.h | 7 +++++--
[...]

PR #68, opened 2024-09-27
u-boot: send hardening warning messages to stderr

rborn-tx:
Send most of the hardening warning/error messages to U-Boot's stderr to increase the chances they are seen.

Related-to: https://github.com/toradex/meta-toradex-security/issues/64

### DIFF
--- a/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch
+++ b/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch
@@ -995,21 +995,21 @@ index 00000000000..462df6709d2
 +				    int argc, char *const argv[],
 +				    char *const reason)
 +{
-+	puts("## WARNING: Command execution ");
-+	puts(simulated ? "WOULD BE DENIED in closed state" : "denied");
-+	puts(" (");
-+	puts(reason);
-+	puts(") for `");
++	eputs("## WARNING: Command execution ");
++	eputs(simulated ? "WOULD BE DENIED in closed state" : "denied");
++	eputs(" (");
++	eputs(reason);
++	eputs(") for `");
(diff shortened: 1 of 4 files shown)


jsrc27:
Do you think the following warning/error messages also need to be sent to stderr?:
https://github.com/toradex/meta-toradex-security/blob/2c5ff367dc53d4ac43cd29e90583ef4e0e94dad8/recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch#L1093

https://github.com/toradex/meta-toradex-security/blob/2c5ff367dc53d4ac43cd29e90583ef4e0e94dad8/recipes-bsp/u-boot/files/0004-toradex-integrate-bootm-protection-downstream.patch#L128 (same on upstream)

https://github.com/toradex/meta-toradex-security/blob/2c5ff367dc53d4ac43cd29e90583ef4e0e94dad8/recipes-bsp/u-boot/files/0004-toradex-integrate-bootm-protection-downstream.patch#L103 (same on upstream)

Or are these different compared to the messages changed in this PR?

rborn-tx:
> Do you think the following warning/error messages also need to be sent to stderr?:

@jsrc27 Good question!

Basically in this PR I chose to send to stderr only messages starting with a double hash mark; those are the messages I considered (during development) as being important and more likely to happen (of course based on personal guess).

The messages you highlighted of course are important (same as hundreds of other errors U-Boot might detect, unrelated to the hardening) but they should be detected during development in which case I'd expect nobody to be executing U-Boot with stdout redirected to nulldev for example.

jsrc27:
@rborn-tx Okay that makes sense just wanted to check/confirm. No other comments from my side then.

jsrc27 (approved):

sergioprado (approved):
LGTM

PR #69, opened 2024-09-27
u-boot: send hardening warning messages to stderr

rborn-tx:
Send most of the hardening warning/error messages to U-Boot's stderr to increase the chances they are seen.

Related-to: https://github.com/toradex/meta-toradex-security/issues/64


rborn-tx:
Guys, I'll go ahead and merge this since it's just the same as with kirkstone which was already merged.

PR #70, opened 2024-09-28
Fix build issue on imx8 (detected on apalis-imx8 and colibri-imx8x)

rborn-tx:
Upstream project meta-freescale has dropped variable BOOT_NAME since c30f12b8 (imx-boot: Drop un-necessary variable BOOT_NAME) which was actually used by this layer to determine the name of the bootloader binary to be signed. Here we follow along by hard-coding the value that was previously in that variable.

See https://github.com/Freescale/meta-freescale/commit/c30f12b809a8cf36043b42c67dd8a11f69d9cf77


sergioprado (approved):
LGTM

PR #73, opened 2024-10-16
Adapt hardening to AM62 (scarthgap)

rborn-tx:
See commit messages for details.

Note: the implementation was tested on an "open" device only.

### DIFF
--- a/classes/tdx-signed-fit-image.inc
+++ b/classes/tdx-signed-fit-image.inc
@@ -17,6 +17,7 @@ FIT_KEY_REQ_ARGS ?= "-batch -new"
 FIT_KEY_SIGN_PKCS ?= "-x509"
 
 # parameters to sign FIT images
+FIT_HASH_ALG ?= "sha256"
 FIT_SIGN_ALG ?= "rsa2048"
 FIT_SIGN_NUMBITS ?= "2048"
 FIT_SIGN_INDIVIDUAL = "0"
--- a/classes/tdx-signed-fit-image.inc
+++ b/classes/tdx-signed-fit-image.inc
@@ -21,3 +21,14 @@ FIT_HASH_ALG ?= "sha256"
 FIT_SIGN_ALG ?= "rsa2048"
 FIT_SIGN_NUMBITS ?= "2048"
 FIT_SIGN_INDIVIDUAL = "0"
+
+# stronger assignments to override settings in meta-ti
+#
+# NOTICE: a side effect of these assignments is that users will need to use
+#         :forcevariable (or an override stronger than :k3) to set them.
+UBOOT_SIGN_KEYDIR:k3 ?= "${TOPDIR}/keys/fit"
+UBOOT_SIGN_KEYNAME:k3 ?= "dev"
+UBOOT_SIGN_IMG_KEYNAME:k3 ?= "dev2"
+FIT_HASH_ALG:k3 ?= "sha256"
+FIT_SIGN_ALG:k3 ?= "rsa2048"
+FIT_SIGN_NUMBITS:k3 ?= "2048"
--- a/classes/tdx-signed-harden.inc
+++ b/classes/tdx-signed-harden.inc
@@ -40,6 +40,7 @@ TDX_SECBOOT_REQUIRED_BOOTARGS:apalis-imx8 ?= "ro rootwait console=tty1 console=t
 TDX_SECBOOT_REQUIRED_BOOTARGS:verdin-imx8mm ?= "ro rootwait console=tty1 console=ttymxc0,115200"
 TDX_SECBOOT_REQUIRED_BOOTARGS:verdin-imx8mp ?= "ro rootwait console=tty1 console=ttymxc2,115200"
 TDX_SECBOOT_REQUIRED_BOOTARGS:colibri-imx8x ?= "ro rootwait console=tty1 console=ttyLP3,115200"
+TDX_SECBOOT_REQUIRED_BOOTARGS:verdin-am62 ?= "ro rootwait console=tty1 console=ttyS2,115200"
 
 # Name of the overlay file (without extension) that will contain the fixed part
 # of the kernel command line; this will be stored inside the FIT image; notice
--- a/classes/tdx-signed-harden.inc
+++ b/classes/tdx-signed-harden.inc
@@ -1,8 +1,8 @@
 # Enable the Secure Boot hardening of U-Boot by Toradex.
 #
-# Currently the hardening implementation requires HAB/AHAB to be enabled so that
-# the combination TDX_IMX_HAB_ENABLE="0" TDX_UBOOT_HARDENING_ENABLE="1" is not
-# valid.
+# Currently the hardening implementation requires HAB/AHAB (or equivalent
+# technology in TI SoCs) to be enabled so that the combination
+# TDX_IMX_HAB_ENABLE="0" and TDX_UBOOT_HARDENING_ENABLE="1" is not valid.
 #
 # Also the hardening itself is supposed to protect the second link of the chain
 # of trust i.e. the link between bootloader and the kernel artifacts. So, the
@@ -13,12 +13,18 @@
 # UBOOT_SIGN_ENABLE are set, by default.
 #
 def default_tdx_uboot_hardening_enable(d):
-    if (bb.utils.to_boolean(d.getVar('TDX_IMX_HAB_ENABLE')) and
+    if ('imx-generic-bsp' in d.getVar('OVERRIDES').split(':') and
+        bb.utils.to_boolean(d.getVar('TDX_IMX_HAB_ENABLE')) and
+        bb.utils.to_boolean(d.getVar('UBOOT_SIGN_ENABLE'))):
+        return True
+    if ('k3' in d.getVar('OVERRIDES').split(':') and
+        bb.utils.to_boolean(d.getVar('TDX_K3_HSSE_ENABLE')) and
         bb.utils.to_boolean(d.getVar('UBOOT_SIGN_ENABLE'))):
         return True
     return False
 
 TDX_UBOOT_HARDENING_ENABLE ?= "${@'1' if default_tdx_uboot_hardening_enable(d) else '0'}"
+TDX_UBOOT_HARDENING_ENABLE:verdin-am62-k3r5 = "0"
 
 # Configure the "kernel command-line protection"; with this protection, the
 # fixed part of the bootargs are saved into the FIT image and checked against
--- a/docs/README-secure-boot.md
+++ b/docs/README-secure-boot.md
@@ -113,6 +113,10 @@ A few variables can be used to configure this feature, including:
 
 The complete list of variables can be found in the `tdx-signed-fit-image.inc` file.
 
+### Configuring FIT image signing / known issues
+
+- On the **Verdin AM62** SoM, some of the configuration variables (e.g. `UBOOT_SIGN_KEYDIR`, `UBOOT_SIGN_KEYNAME` (check the complete list in `tdx-signed-fit-image.inc`)) are set through override `k3` to ensure the values coming from layer `meta-toradex-security` override those from layer `meta-ti-bsp`. Due to this, the recommended way to set those variables is via override `forcevariable`.
+
 ## Configuring rootfs image signing
 
 When the `tdxref-signed` class is inherited, the rootfs image will be generated using the `dm-verity` kernel feature.
--- a/docs/README-secure-boot.md
+++ b/docs/README-secure-boot.md
@@ -65,9 +65,11 @@ The hardening features above are controlled by the following variables:
 
 | Variable | Description | Default value |
 | :------- | :---------- | :------------ |
-| `TDX_UBOOT_HARDENING_ENABLE` | Enable hardening features as a whole | `1` if both `TDX_IMX_HAB_ENABLE` and `UBOOT_SIGN_ENABLE` are set; `0` otherwise |
+| `TDX_UBOOT_HARDENING_ENABLE` | Enable hardening features as a whole | `1` if both secure boot support (controlled by variable `TDX_IMX_HAB_ENABLE` (NXP) or `TDX_K3_HSSE_ENABLE` (TI)) and FIT signing (controlled by `UBOOT_SIGN_ENABLE`) are enabled; `0` otherwise |
 | `TDX_SECBOOT_REQUIRED_BOOTARGS` | Expected value for the fixed part of the kernel command line | Different value for each machine |
 | `TDX_AMEND_BOOT_SCRIPT` | When set to `1` the boot script will be amended to make it suitable for secure boot; this only works with the script provided by Toradex for BSP reference images; users employing a custom script should set this to `0` | Same value as variable `TDX_UBOOT_HARDENING_ENABLE` |
+
+### U-Boot hardening / setup
 
 The behavior of the different hardening features can be set via the control FDT (see [Devicetree Control in U-Boot](https://u-boot.readthedocs.io/en/stable/develop/devicetree/control.html)). Setting the control FDT at build time can be achieved by adding extra device-tree [.dtsi fragments](https://u-boot.readthedocs.io/en/stable/develop/devicetree/control.html#external-dtsi-fragments) to U-Boot and setting the Kconfig variable `CONFIG_DEVICE_TREE_INCLUDES` appropriately; with Yocto/OE this would normally involve adding small patches to U-Boot and appending changes to its recipe but the details are outside the scope of the present document.
 
@@ -94,6 +96,10 @@ The following device-tree fragment shows all the nodes and properties that can b
 The command categories are currently only available as part of a [patch](./recipes-bsp/u-boot/files/0001-toradex-common-add-command-whitelisting-modules.patch) in header `cmd-categories.h`. The default FDT is part of another [patch](./recipes-bsp/u-boot/files/0002-toradex-dts-add-fragment-file-to-configure-secure-bo.patch) in file `tdx-secboot.dtsi`.
 
 <!-- TODO: Make more user-friendly instructions on setting the control FDT. -->
+
+### U-Boot hardening / known issues
+
+- On the **Verdin AM62** SoM, the hardening does not cover the bootloader running on the R5 processor (the boot master); we have plans to evaluate the need for such a protection and implementing it if actually needed.
 
 ## Configuring FIT image signing
 
@@ -136,7 +142,3 @@ When `tdxref-signed` is used to enable secure boot, the rootfs image is generate
 Because `dm-verity` is read-only, you might want to create an additional partition in the eMMC to store persistent data.
 
 If that is the case, you can use the `tdx-tezi-data-partition` class. For more information, have a look at its documentation ([README-data-partition.md](README-data-partition.md)).
-
-## Known issues and limitations
-
-- Currently the hardening is implemented/integrated on NXP SoCs only; when building for other SoC vendors one has to disable the feature (set `TDX_UBOOT_HARDENING_ENABLE=` to `0`) or the build will fail.
--- a/recipes-bsp/u-boot/u-boot-harden.inc
+++ b/recipes-bsp/u-boot/u-boot-harden.inc
@@ -6,6 +6,8 @@ TDX_UBOOT_HARDENING_PATCHES_DOWNSTREAM = "\
     file://0005-toradex-integrate-CLI-access-protection-downstream.patch \
     file://0006-toradex-add-implementation-of-bootargs-protection.patch \
     file://0007-toradex-integrate-bootargs-protection-downstream.patch \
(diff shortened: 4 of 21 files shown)


rborn-tx:
@lucas-akira tagging you to keep you informed...

docs/README-secure-boot.md
```diff
@@ -113,6 +115,10 @@ A few variables can be used to configure this feature, including:
 
 The complete list of variables can be found in the `tdx-signed-fit-image.inc` file.
 
+### Configuring FIT image signing / known issues
+
+- On the **Verdin AM62** SoM, some of the configuration variables (e.g. `UBOOT_SIGN_KEYDIR`, `UBOOT_SIGN_KEYNAME` (check the complete list in `tdx-signed-fit-image.inc`)) are set through override `k3` to ensure the values coming from layer `meta-toradex-security` override those from layer `meta-ti-bsp`; this should be a temporary measure until fixes to the latter are upstreamed. Due to this, the recommended way to set those variables at the moment is via override `forcevariable`.
+
```
sergioprado, line 125: Since this is unrelated to the hardening feature, can it be added as a separate patch? This will also help later to revert it as soon as it is fixed upstream.

recipes-bsp/u-boot/u-boot-harden.inc
```diff
@@ -39,8 +39,24 @@ SRC_URI:append = "\
...
+                'the whitelisting feature (part of the hardening) currently relies on K3.'
+    fi
+}
+
+do_compile:prepend:k3() {
+    k3_compile_sanity_checks
+}
+
+do_compile:prepend:k3r5() {
+    k3_compile_sanity_checks
```
sergioprado: Is it required to call this function also for `k3r5`?

classes/tdx-signed-harden.inc
```diff
@@ -13,12 +13,18 @@
 # UBOOT_SIGN_ENABLE are set, by default.
 #
 def default_tdx_uboot_hardening_enable(d):
-    if (bb.utils.to_boolean(d.getVar('TDX_IMX_HAB_ENABLE')) and
+    if ('imx-generic-bsp' in d.getVar('MACHINEOVERRIDES_EXTENDER').split(':') and
```
sergioprado: Why not `OVERRIDES` instead of `MACHINEOVERRIDES_EXTENDER`?

classes/tdx-signed-fit-image.inc
```diff
@@ -21,3 +21,14 @@ FIT_HASH_ALG ?= "sha256"
...
+# TODO: Get rid of this block of assignments involving :k3 once fixes to meta-ti are upstreamed.
+# NOTE: A side effect of these assignments is that users will need to use :forcevariable to set them.
+#
+# stronger assignments to override settings in meta-ti
+UBOOT_SIGN_KEYDIR:k3 ?= "${TOPDIR}/keys/fit"
+UBOOT_SIGN_KEYNAME:k3 ?= "dev"
+UBOOT_SIGN_IMG_KEYNAME:k3 ?= "dev2"
+FIT_HASH_ALG:k3 ?= "sha256"
+FIT_SIGN_ALG:k3 ?= "rsa2048"
+FIT_SIGN_NUMBITS:k3 ?= "2048"
```
sergioprado, line 34: Can you explain better what was the build error you had without setting the FIT_SIGN_NUMBITS variable? Do you have the logs to share? I don't remember having errors when building in my environment without this variable. I just want to understand why you have the error in your environment and I don't.

sergioprado (commented):
Looks good in general. I just have a few comments.

docs/README-secure-boot.md
sergioprado: Maybe this can be done together with https://github.com/toradex/meta-toradex-security/pull/73/commits/d39f38a44fa462af7b914d832e01c6f4869f0018, so it can be easily reverted later?

jsrc27:
Think the table here: https://github.com/toradex/meta-toradex-security/blob/kirkstone-6.x.y/docs/README-secure-boot.md?plain=1#L68

Needs to also be updated to reflect the TI specific `TDX_K3_HSSE_ENABLE` variable. To clarify the table should at last mention the interaction with the TI variable as well.

classes/tdx-signed-harden.inc
```diff
@@ -13,12 +13,18 @@
...
+        bb.utils.to_boolean(d.getVar('UBOOT_SIGN_ENABLE'))):
+        return True
+    if ('k3' in d.getVar('OVERRIDES').split(':') and
+        bb.utils.to_boolean(d.getVar('TDX_K3_HSSE_ENABLE')) and
         bb.utils.to_boolean(d.getVar('UBOOT_SIGN_ENABLE'))):
         return True
     return False
 
 TDX_UBOOT_HARDENING_ENABLE ?= "${@'1' if default_tdx_uboot_hardening_enable(d) else '0'}"
+TDX_UBOOT_HARDENING_ENABLE:verdin-am62-k3r5 = "0"
```
jsrc27, line 27: Out of curiosity I know this refers to the "other" machine configuration for the am62 and not the "real" one of `verdin-am62`. 

I assume you explicitly set this to 0 here for the "other" configuration due to weird interactions?

recipes-bsp/u-boot/u-boot-harden.inc
```diff
@@ -16,19 +33,39 @@ TDX_UBOOT_HARDENING_PATCHES_UPSTREAM = "\
...
-        bbfatal 'The combination TDX_IMX_HAB_ENABLE = "0" and TDX_UBOOT_HARDENING_ENABLE = "1" is not allowed: the whitelisting feature (part of the hardening) currently relies on HAB/AHAB.'
+        bbfatal 'The combination TDX_IMX_HAB_ENABLE="0" and TDX_UBOOT_HARDENING_ENABLE="1" is not allowed:' \
+                'the whitelisting feature (part of the hardening) currently relies on HAB/AHAB.'
     fi
 }
+
+k3_compile_sanity_checks() {
+    if [ "${TDX_K3_HSSE_ENABLE}" = "0" ] && [ "${TDX_UBOOT_HARDENING_ENABLE}" = "1" ]; then
+        bbfatal 'The combination TDX_K3_HSSE_ENABLE="0" and TDX_UBOOT_HARDENING_ENABLE="1" is not allowed:' \
+                'the whitelisting feature (part of the hardening) currently relies on K3.'
```
jsrc27, line 61: Is saying "K3" here in the message the correct terminology? Isn't K3 more referring to the machine architecture. I think HSSE would be the equivalent term to HAB/AHAB, right?

jsrc27 (commented):
No major comments from me just some minor questions/comments here and there.

rborn-tx:
Moved to draft to fix an issue I didn't notice before.

docs/README-secure-boot.md
rborn-tx: Given the status of our upstream patch, we'll probably need to review our position here. We could make this a more definitive solution with a known issue or we could consider that idea of not setting the signing-related variables in our layer.

recipes-bsp/u-boot/u-boot-harden.inc
rborn-tx: Yes, I believe so. However, now I realize the check should be different between the two cases (k3, k3r5). In case of k3r5 the hardening cannot be enabled at all at the moment. This might leave some attack surface uncovered but to solve it I'll need to better understand the architecture and what tasks the R5 processor is responsible for.

classes/tdx-signed-harden.inc
rborn-tx: It's like this by mistake. Thanks for catching it.

classes/tdx-signed-fit-image.inc
rborn-tx: IIRC, I got the exact same error that you reported in our internal ticket (TOR-3556), namely:

```
|   Timestamp:    Thu Jan  1 00:00:00 1970
| Signature written to '/home/sprado/projects/toradex/build/torizon/scarthgap/build-verdin-am62/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/build/fitImage-linux', node '/configurations/conf-verd
in-am62_spidev_overlay.dtbo/signature-1'                                                                                                                                                                           
| Public key written to 'u-boot.dtb', node '/signature/key-custMpk'
| Signature check bad (error 1)                                                                          
| Verifying Hash Integrity for node 'conf-ti_k3-am625-verdin-nonwifi-dahlia.dtb'... sha512,rsa4096:custMpk-
|  error!                                                                                                
| Verification failed for '(null)' hash node in 'conf-ti_k3-am625-verdin-nonwifi-dahlia.dtb' config node 
| Failed to verify required signature 'key-custMpk'                                                      
| WARNING: exit code 1 from a shell command.                                                             
ERROR: Task (/home/sprado/projects/toradex/build/torizon/scarthgap/build-verdin-am62/conf/../../layers/meta-toradex-ti/recipes-bsp/u-boot/u-boot-toradex-ti_2024.04.bb:do_uboot_assemble_fitimage) failed with exit
 code '1'
```

And I ran a couple of build tests where I found the error happened when `FIT_SIGN_ALG` and `FIT_SIGN_NUMBITS` were not aligned.

Anyways, I'll re-run some tests to confirm and report back here...

rborn-tx:
> Think the table here: https://github.com/toradex/meta-toradex-security/blob/kirkstone-6.x.y/docs/README-secure-boot.md?plain=1#L68
> 
> Needs to also be updated to reflect the TI specific `TDX_K3_HSSE_ENABLE` variable. To clarify the table should at last mention the interaction with the TI variable as well.

Good point. I'll do that.

recipes-bsp/u-boot/u-boot-harden.inc
rborn-tx: I'm not 100% sure about the terminology. Maybe @sergioprado could help here. This [page](https://software-dl.ti.com/jacinto7/esd/processor-sdk-rtos-jacinto7/latest/exports/docs/psdk_rtos/docs/user_guide/developer_notes_security_build.html) refers to K3 as a security architecture and to HS-FS/HS-SE as states of a HS device.

If it helps, let me clarify the dependency we have: basically we need a function provided by module `arch/arm/mach-k3/common.c` in the U-Boot source code which is specific to the K3 architecture. Btw, that was the main reason I used the word "K3" in the fatal message.

classes/tdx-signed-harden.inc
rborn-tx: I didn't want the hardening enabled when compiling U-Boot for that other machine simply because it wasn't designed for it. As I mentioned in another message, this might leave some uncovered attack surface. If other protections are needed, I imagine they would be different from what we have today and it would require a new implementation. I intended to capture this in a separate ticket.

classes/tdx-signed-fit-image.inc
rborn-tx: Ok,

1. Removing any assignments to `FIT_HASH_ALG`, `FIT_SIGN_ALG` and `FIT_SIGN_NUMBITS` from our layer (the normal ones and the ones with the `k3` override) the resolved values of these variables becomes:

```
FIT_HASH_ALG = "sha512"
FIT_SIGN_ALG = "rsa4096"
FIT_SIGN_NUMBITS = <not-set> # when baking virtual/bootloader
FIT_SIGN_NUMBITS = "2048"    # when baking virtual/kernel
```

In this case, the build fails with pretty much the same error (except for the name of the key, whose assignment I kept in the layer):

```
| Signature written to '/workdir/build-verdin-am62-sec/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/build/fitImage-linux', node '/configurations/conf-verdin-am62_spidev_overlay.dtbo/signature-1'
| Public key written to 'u-boot.dtb', node '/signature/key-dev'
| Signature check bad (error 1)
| Verifying Hash Integrity for node 'conf-ti_k3-am625-verdin-nonwifi-dahlia.dtb'... sha512,rsa4096:dev-
|  error!
| Verification failed for '(null)' hash node in 'conf-ti_k3-am625-verdin-nonwifi-dahlia.dtb' config node
| Failed to verify required signature 'key-dev'
| WARNING: /workdir/build-verdin-am62-sec/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/temp/run.do_uboot_assemble_fitimage.1310264:213 exit 1 from 'uboot-fit_check_sign -k "u-boot.dtb" -f /workdir/build-verdin-am62-sec/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/build/fitImage-linux'
| WARNING: Backtrace (BB generated script):
|       #1: concat_dtb, /workdir/build-verdin-am62-sec/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/temp/run.do_uboot_assemble_fitimage.1310264, line 213
|       #2: uboot_assemble_fitimage_helper, /workdir/build-verdin-am62-sec/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/temp/run.do_uboot_assemble_fitimage.1310264, line 186
|       #3: do_uboot_assemble_fitimage, /workdir/build-verdin-am62-sec/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/temp/run.do_uboot_assemble_fitimage.1310264, line 172
|       #4: main, /workdir/build-verdin-am62-sec/tmp/work/verdin_am62-tdx-linux/u-boot-toradex-ti/2024.04+git/temp/run.do_uboot_assemble_fitimage.1310264, line 391
ERROR: Task (/workdir/build-verdin-am62-sec/conf/../../layers/meta-toradex-ti/recipes-bsp/u-boot/u-boot-toradex-ti_2024.04.bb:do_uboot_assemble_fitimage) failed with exit code '1'
```

2. Repeating the same test but adding only the assignment `FIT_SIGN_NUMBITS="4096"` (in order to match the setting `FIT_SIGN_ALG="rsa4096"` coming from meta-ti) the build succeeds and the image boots normally.

From this I think we can conclude that the issue is just the mismatch between `FIT_SIGN_ALG` and `FIT_SIGN_NUMBITS` caused by the setting of only `FIT_SIGN_ALG` in `meta-ti` which I believe is the thing we need to fix upstream.

docs/README-secure-boot.md
sergioprado: To keep consistency between all SoC platforms, I would prefer to have the defaults configured in our layer (using an override) and then documenting how the user can override it via the `forcevariable` override.

recipes-bsp/u-boot/u-boot-harden.inc
sergioprado: I think U-Boot running on the R5 processor doesn't expose any surface attack (it doesn't have a serial port console, it doesn't use external files for configuration, etc). But I am not 100% sure, so makes sense to do some investigation.

classes/tdx-signed-fit-image.inc
sergioprado: Gotcha. Thanks for testing!

recipes-bsp/u-boot/u-boot-harden.inc
sergioprado: K3 stands for Keystone 3 and refers to a the family of SoCs from TI. As far as I could tell, there is no name for the secure boot implementation on TI chips (as we have HAB/AHAB for NXP chips). So I think it is fair to use K3 here (I have used K3 in variables and in the secure boot documentation as well).

docs/README-secure-boot.md
rborn-tx: That sounds sensible to me. The latest version is following that approach already. The metadata changes and corresponding documentation changes were placed into the two sequential commits:

- [97d7a00](https://github.com/toradex/meta-toradex-security/pull/73/commits/97d7a004d5910308d7bf9bc2a22ecc6851864b33) README-secure-boot: list known issues for TI (verdin-am62)
- [e7009cf](https://github.com/toradex/meta-toradex-security/pull/73/commits/e7009cf11a7e58a60e91936d6929a0d1b2dd3f4d) tdx-signed-fit-image: override signing parameters from TI

just in case we ever want to revert this, even though this is unlikely to happen.

recipes-bsp/u-boot/u-boot-harden.inc
rborn-tx: Latest version has been updated to have different checks for k3 and k3r5 and as follow up I'm going to create a ticket for further investigations. Nonetheless I've added a note to the documentation saying that we have plans to do that investigation for the sake of transparency:

[9c92d1b](https://github.com/toradex/meta-toradex-security/pull/73/commits/9c92d1ba9ab79a29ef2f822993866a08099933aa) README-secure-boot: update documentation

classes/tdx-signed-harden.inc
rborn-tx: Fixed. Solving this thread...

recipes-bsp/u-boot/u-boot-harden.inc
rborn-tx: @jsrc27 Thanks for raising the question so we could align our understanding of the topic.
@sergioprado Thanks for the clarification.

I'll go ahead and close this thread then.

classes/tdx-signed-harden.inc
rborn-tx: @jsrc27 Is it okay to resolve this thread?

classes/tdx-signed-fit-image.inc
```diff
@@ -21,3 +21,14 @@ FIT_HASH_ALG ?= "sha256"
...
+# stronger assignments to override settings in meta-ti
+#
+# NOTICE: a side effect of these assignments is that users will need to use
+#         :forcevariable (or an override stronger than :k3) to set them.
+UBOOT_SIGN_KEYDIR:k3 ?= "${TOPDIR}/keys/fit"
+UBOOT_SIGN_KEYNAME:k3 ?= "dev"
+UBOOT_SIGN_IMG_KEYNAME:k3 ?= "dev2"
+FIT_HASH_ALG:k3 ?= "sha256"
+FIT_SIGN_ALG:k3 ?= "rsa2048"
+FIT_SIGN_NUMBITS:k3 ?= "2048"
```
sergioprado, line 34: Alright. So I had an idea that might work and fix this issue without any side efffect. What if we create a custom variable for each FIT variable that we want to set as default. Something like this:

```
TDX_UBOOT_SIGN_ENABLE ?= "${TOPDIR}/keys/fit"
UBOOT_SIGN_KEYDIR = "${TDX_UBOOT_SIGN_ENABLE}"

TDX_UBOOT_SIGN_KEYNAME ?= "dev"
UBOOT_SIGN_KEYNAME = "${TDX_UBOOT_SIGN_KEYNAME}"

TDX_UBOOT_SIGN_IMG_KEYNAME ?= "dev2"
UBOOT_SIGN_IMG_KEYNAME = "${TDX_UBOOT_SIGN_IMG_KEYNAME}"
```

With this:

- We can have our defaults without worring about being override by a machine.
- The user can easily override it without using forcevariable.
- We don't duplicate the same value indifferent variables (one for iMX and one for K3), making it easy to maintain.

This would also mean we would need to document those variables.

Makes sense?

rborn-tx: That's a great idea. Thanks! I'll go ahead and do it.

rborn-tx: @sergioprado When starting to make the changes I thought of an issue that made me stop: if we follow this approach, we'd be overriding settings done by customers at the moment, i.e. it would be a breaking change, right?

Then I further thought of some other points against it:

1. In the end we'd be doing this just to compensate for an issue that only happens when the TI layer is in use; at the moment we have one machine using that layer which means the decision would be based on something (I hope) is an exception case.
2. We don't implement the FIT image signing in our layer, but with the proposed change we'd be adding a group of variables to configure the signing that is provided by another layer; however, we wouldn't override each and every configuration variable which means a user would need to always check both our documentation/code and openembedded's to figure out how to appropriately apply the settings they want. Personally, I'd prefer that users resort mainly to openembedded docs.
3. Though I understand that having our defaults looks good at first sight, I'm not sure if we should override settings from the layer providing the feature unless strictly necessary for our functionalities to work; I say that because I imagine the providers of the feature should only change the defaults when there is a very good reason for it and they should know better than us how to best configure their feature. Of course the same statement holds for meta-ti. A question to comes to mind is: why is our defaults better than the ones from meta-ti and why are the meta-ti ones better than the ones from openembedded?
4. And let me just repeat: this would be a breaking change from NXP users perspective.

Anyways, we should probably talk to avoid a huge thread on this topic :slightly_smiling_face:...

sergioprado: Yeah, that is really unfortunate. Alright, to not block this PR, lets leave it as is right now and maybe later we re-evaluate the situation.

sergioprado (approved):
@rborn-tx LGTM. Please let me know if you need some help to test this on a closed device before merging it.

Will also wait for @jsrc27 approvals.

jsrc27 (approved):
LGTM as well

PR #74, opened 2024-10-16
Adapt hardening to AM62 (kirkstone)

rborn-tx:
Same as https://github.com/toradex/meta-toradex-security/pull/73 but for kirkstone.

Though kirkstone is entering maintenance mode, I ended up making this adaptation also in it just to avoid having a gap on the hardening feature just on the AM62 machine.

Note: the implementation was tested on an "open" device only.


rborn-tx:
Merging this since the same changes were reviewed and merged on scarthgap already.

PR #79, opened 2024-10-24
Reorganize hardening patches

rborn-tx:
To simplify maintenance, split patches into common and specific ones. The common ones are supposed to be exactly the same no matter for what branch of U-Boot they are applied to (because they add/change files that were not originally present in U-Boot's code) and the specific ones are those involving some integration with U-Boot's code and as such they will likely need changes over time to align with the changes to U-Boot. In other words, only the specific ones will ever need rebases when U-Boot is updated.

Additionally, fix the path to some patch files as referenced by the documentation.


jsrc27 (approved):
LGTM

sergioprado (approved):
LGTM

PR #83, opened 2024-11-13
Allow tdx-encrypted to be used with Torizon (plus improvements)

rborn-tx:
This should fix the problems with using `tdx-encrypted` with Torizon OS - this is achieved by the first two commits.

The other commits introduce a few other improvements (and a bugfix). Please refer to the commit messages for details.

### DIFF
--- a/classes/tdx-encrypted.bbclass
+++ b/classes/tdx-encrypted.bbclass
@@ -1,5 +1,5 @@
-# Enable encryption support
-TDX_ENC_ENABLE = "1"
+# override for conditional assignment
+DISTROOVERRIDES .= ":tdx-encrypted"
 
 # Encryption key backend
 # This variable defines how the encryption key is managed
--- a/classes/tdx-encrypted.bbclass
+++ b/classes/tdx-encrypted.bbclass
@@ -41,6 +41,13 @@ TDX_ENC_STORAGE_RESERVE ?= "0"
 # Defines where the encrypted storage will be mounted
 TDX_ENC_STORAGE_MOUNTPOINT ?= "/run/encdata"
 
+# Extra arguments passed to "mkfs" when creating the filesystem on top
+# of the encrypted storage
+TDX_ENC_STORAGE_MKFS_ARGS ?= ""
+
+# Extra arguments passed to "mount" when mounting the encrypted storage
+TDX_ENC_STORAGE_MOUNT_ARGS ?= ""
+
 # Enables preservation of existing data on encrypted device
 TDX_ENC_PRESERVE_DATA ?= "0"
 
--- a/classes/tdx-tezi-data-partition.bbclass
+++ b/classes/tdx-tezi-data-partition.bbclass
@@ -1,5 +1,5 @@
 # override for conditional assignment
-DISTROOVERRIDES:append = ":tdx-tezi-data-partition"
+DISTROOVERRIDES .= ":tdx-tezi-data-partition"
 
 # data partition filesystem type
 # supported values: ext2, ext3, ext4, fat, ubifs
--- a/classes/tdx-tezi-data-partition.bbclass
+++ b/classes/tdx-tezi-data-partition.bbclass
@@ -9,8 +9,18 @@ TDX_TEZI_DATA_PARTITION_TYPE ?= "ext4"
 # data partition label
 TDX_TEZI_DATA_PARTITION_LABEL ?= "DATA"
 
-# automount data partition at boot time
-TDX_TEZI_DATA_PARTITION_AUTOMOUNT ?= "1"
+# automount data partition at boot time; possible values:
+#  "-1": do not change fstab
+#   "0": add entry to fstab with the noauto option
+#   "1": add entry to fstab with the auto option
+TDX_TEZI_DATA_PARTITION_AUTOMOUNT ?= "${TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT}"
+
+# define the default value of above variable based on tdx-encrypted being
+# in use or not; notice TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT is for
+# internal usage of the present class and it should not be modified by
+# users (who should only care about TDX_TEZI_DATA_PARTITION_AUTOMOUNT)
+TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT = "1"
+TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT:tdx-encrypted = "-1"
 
 # data partition mount point
 TDX_TEZI_DATA_PARTITION_MOUNTPOINT ?= "/data"
--- a/docs/README-encryption.md
+++ b/docs/README-encryption.md
@@ -63,19 +63,24 @@ TDX_ENC_KEY_BACKEND:forcevariable = "tpm"
 
 Make sure to use the `forcevariable` override, so your configuration takes precedence over the default one.
 
-A few additional variables are available to customize the behavior of the data-at-rest encryption feature. Here is the complete list:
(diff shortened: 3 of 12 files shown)


rborn-tx:
I imagine we'll want to backport this to kirkstone. Opinions?

jsrc27 (approved):
LGTM

As for bringing this to Kirkstone, if there's no additional complexities then I support this plan.

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
```diff
@@ -343,8 +343,10 @@ tdx_enc_clear_keys_keyring() {
 
 # umount partition
 tdx_enc_partition_umount() {
-    tdx_enc_log "Unmounting dm-crypt partition..."
-    umount "${TDX_ENC_STORAGE_MOUNTPOINT}"
+    for mnt in $(lsblk /dev/mapper/"${TDX_ENC_DM_DEVICE}" -n -o MOUNTPOINTS); do
```
sergioprado, line 352: Change here looks good. Just a small typo in the commit message: `paritition` -> `partition`.

classes/tdx-encrypted.bbclass
```diff
@@ -41,6 +41,13 @@ TDX_ENC_STORAGE_RESERVE ?= "0"
 # Defines where the encrypted storage will be mounted
 TDX_ENC_STORAGE_MOUNTPOINT ?= "/run/encdata"
 
+# Extra arguments passed to "mkfs" when creating the filesystem on top
+# of the encrypted storage
+TDX_ENC_STORAGE_MKFS_ARGS ?= ""
+
+# Extra arguments passed to "mount" when mounting the encrypted storage
+TDX_ENC_STORAGE_MOUNT_ARGS ?= ""
```
sergioprado, line 49: Nice addition! Maybe I am missing something, but did you added those variables to the documentation? If not, would you do it please?

classes/tdx-encrypted.bbclass
```diff
@@ -24,12 +24,6 @@ TDX_ENC_KEY_DIR ?= "/var/local/private/.keys"
 # encryption key blob file name
 TDX_ENC_KEY_FILE ?= "tdx-enc-key.blob"
 
-# Type of encryption
-# This variable defines what will be encrypted
-# Available options:
-#    partition -> encrypt a full partition
-TDX_ENC_STORAGE_TYPE ?= "partition"
```
sergioprado: Any particular reason to remove this variable, besides not being used for now? My initial idea was to support other types of encryption, like `file` based encryption. Hopefully I will be able to implement that soon, so if you don't mind, I would like to keep it. :)

classes/tdx-tezi-data-partition.bbclass
```diff
@@ -9,8 +9,15 @@ TDX_TEZI_DATA_PARTITION_TYPE ?= "ext4"
...
-TDX_TEZI_DATA_PARTITION_AUTOMOUNT ?= "1"
+# automount data partition at boot time; possible values:
+#  "-1": do not change fstab
+#   "0": add entry to fstab with the noauto option
+#   "1": add entry to fstab with the auto option
+TDX_TEZI_DATA_PARTITION_AUTOMOUNT ?= "${TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT}"
+
+# default value of above variable depends on tdx-encrypted being in use
+TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT = "1"
+TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT:tdx-encrypted = "-1"
```
sergioprado, line 23: If a user looks at this, she might get confused about which variable needs to be set: `TDX_TEZI_DATA_PARTITION_AUTOMOUNT` or `TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT`, although `TDX_TEZI_DATA_PARTITION_AUTOMOUNT` is the one documented.

Makes sense to add a comment to emphasize that `TDX_TEZI_DATA_PARTITION_AUTOMOUNT_DEFAULT` is for internal usage, and the user should set the `TDX_TEZI_DATA_PARTITION_AUTOMOUNT` variable?

docs/README-encryption.md
```diff
@@ -63,19 +63,22 @@ TDX_ENC_KEY_BACKEND:forcevariable = "tpm"
 
 Make sure to use the `forcevariable` override, so your configuration takes precedence over the default one.
 
-A few additional variables are available to customize the behavior of the data-at-rest encryption feature. Here is the complete list:
+A few additional variables are available to customize the behavior of the data-at-rest encryption feature. Here is list of the most important ones (for the full list refer to `tdx-encrypted.bbclass`):
```
sergioprado, line 66: Is there any other user-facing variable not documented here?

sergioprado (commented):
Nice improvements to the encryption feature!

Besides a few comments, just to make sure, did you test this on a BSP image, to make sure there are no regressions?

I would also support cherry-picking this to kirkstone.

Thanks!

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
rborn-tx: Nice catch!

classes/tdx-encrypted.bbclass
rborn-tx: I added that generic message asking users to see the class file for the full list of variables. Before, there was already variables related to the data preservation that were not documented either. What should we do? Should we document all indeed?

Update: anyways I've added the two new variables to the description table in `README-encryption.md`

classes/tdx-encrypted.bbclass
rborn-tx: If we have intentions of using it in the near future I don't see any problem; otherwise I'd add it when we introduce the feature using it. I'll put it back then...

classes/tdx-tezi-data-partition.bbclass
rborn-tx: Yeah, it does make sense.

docs/README-encryption.md
rborn-tx: As I mentioned, the ones related to data preservation. So the question that remains here is whether we want to document all indeed.

rborn-tx:
> Besides a few comments, just to make sure, did you test this on a BSP image, to make sure there are no regressions?

Yes, I forgot to mention: tests were performed on BSP and Torizon OS; also, I've tested the individual usage of the classes (tdx-tezi-data-partition / tdx-encrypted) and both together.

classes/tdx-encrypted.bbclass
rborn-tx: Done

docs/README-encryption.md
rborn-tx: Done

classes/tdx-tezi-data-partition.bbclass
rborn-tx: Done

recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
rborn-tx: Done

classes/tdx-encrypted.bbclass
rborn-tx: Closing this thread since the changes asked are done and leaving the last remaining topic on documenting the data preservation variables to the other open thread.

docs/README-encryption.md
sergioprado: Yeah, you are right. Makes sense to document it. But does not need to be done as part of this PR. Thanks for pointing that out in the documentation.

sergioprado:
LGTM. Merging now.

PR #87, opened 2025-01-17
OP-TEE improvements

sergioprado:
Some improvements related to OP-TEE support. This is part of the work to support OP-TEE on iMX6/7 devices.


jsrc27 (approved):
LGTM. I'll just merge it now since Rogerio is out on vacation.

PR #88, opened 2025-01-29
Add support for using RPMB as a secure storage device

sergioprado:
Tested on Verdin iMX8MP, Verdin iMX8MM and Verdin AM62.

You can view the documentation changes more clearly here: https://github.com/sergioprado/meta-toradex-security/blob/optee-rpmb/docs/README-optee.md

### DIFF
--- a/docs/README-optee.md
+++ b/docs/README-optee.md
@@ -26,27 +26,10 @@ A few variables can be used to customize the behavior of this feature:
 | :------- | :---------- | :------------ |
 | TDX_OPTEE_DEBUG | Enable OP-TEE debug messages to the serial console (`1` to enable or `0` to disable) | `0` |
 | TDX_OPTEE_INSTALL_TESTS | Enable the installation of OP-TEE test applications (`1` to enable or `0` to disable) | `0` |
-| TDX_OPTEE_FTPM | Enable support for a firmware TPM (fTPM) implementation running as trusted application in OP-TEE (`1` to enable or `0` to disable). For more information on how an fTPM works, see the next session. | `0` |
-
-## fTPM support in OP-TEE
-
-Trusted Platform Modules (TPMs) are designed to enhance system security by performing cryptographic operations such as key generation, encryption, and digital signing. Their primary role is to provide secure storage for cryptographic keys and ensure system integrity by measuring the software environment during the boot process.
-
-An fTPM (Firmware-based Trusted Platform Module) is a type of TPM that operates within firmware, as opposed to being a discrete hardware chip. While it offers similar functionality, it runs inside a Trusted Execution Environment (TEE), such as OP-TEE.
-
-The [fTPM implementation](https://github.com/microsoft/ms-tpm-20-ref) integrated into this layer uses the secure storage API provided by OP-TEE. This secure storage currently saves data in `/data/tee/`, where persistent information such as cryptographic keys and other general-purpose data is securely stored. All stored data is encrypted using a Hardware Unique Key (HUK), which is supplied by the CAAM driver on i.MX platforms and the DMSC subsystem on K3 based devices (e.g. AM6X).
-
-For further details on how secure storage works in OP-TEE, refer to the [OP-TEE documentation](https://optee.readthedocs.io/en/latest/architecture/secure_storage.html).
-
-It's important to note that an fTPM may not entirely replace the need for a discrete TPM chip, as this depends on the specific use case and the product's threat model. Nevertheless, an fTPM can provide adequate security for certain scenarios, especially when incorporating a hardware TPM chip is impractical.
-
-## Read-write filesystem for OP-TEE
-
-OP-TEE needs a read-write filesystem (by default, it writes to `/data/tee/`). When rootfs signature checking is enabled via the `tdxref-signed` class, the rootfs image will be generated using the `dm-verity` kernel feature, which is read-only.
-
-In this case, to provide a read-write filesystem to OP-TEE, the `tdx-tezi-data-partition` class will be automatically inherited. This class will create an additional partition in the eMMC and mount it by default at `/data`. For more information on how this class works, have a look at its documentation ([README-data-partition.md](README-data-partition.md)).
-
-In case your rootfs is read-only and you are not using a Toradex Easy Installer image, make sure the `/data` directory is mounted at a read-write partition, or OP-TEE will not work properly.
+| TDX_OPTEE_FS_RPMB | Enable support for using the eMMC RPMB partition as a secure storage device (`1` to enable or `0` to disable) | `0` |
+| TDX_OPTEE_FS_RPMB_DEV_ID | Configure the eMMC RPMB partition device node. For example, if configured with `2`, the TEE supplicant will use `/dev/mmcblk2rpmb` to communicate with to the RPMB partition | `0` |
+| TDX_OPTEE_FS_RPMB_MODE | RPMB secure storage operation mode. Valid options are `development` (to be used during development), `factory` (to enroll the RPMB key in a secure provisioning environment) and `production` (to be used in production). For more information on how to configure this variable, see the session [RPMB support in OP-TEE](/docs/README-optee.md#rpmb-support-in-op-tee) | `development` |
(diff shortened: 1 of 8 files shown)


jsrc27 (approved):
LGTM

docs/README-optee.md
```diff
@@ -26,46 +26,114 @@ A few variables can be used to customize the behavior of this feature:
...
+- The first one relies on the normal world (REE) file system.
+- The second one makes use of the Replay Protected Memory Block (RPMB) partition of an eMMC device.
+
+When relying on the normal world file system, OP-TEE uses `/data/tee/` as its secure storage location in the Linux file system. All stored data is encrypted using a Hardware Unique Key (HUK), which is supplied by the CAAM driver on i.MX platforms and the DMSC subsystem on K3-based devices (e.g., AM6X).
+
+To be able to write and persist data, OP-TEE needs a read-write filesystem mounted at `/data`. When rootfs signature checking is enabled via the `tdxref-signed` class, the rootfs image will be generated using the `dm-verity` kernel feature, which is read-only.
 
 In this case, to provide a read-write filesystem to OP-TEE, the `tdx-tezi-data-partition` class will be automatically inherited. This class will create an additional partition in the eMMC and mount it by default at `/data`. For more information on how this class works, have a look at its documentation ([README-data-partition.md](README-data-partition.md)).
```
rborn-tx: I'd suggest joining these two lines together.

docs/README-optee.md
```diff
@@ -26,46 +26,114 @@ A few variables can be used to customize the behavior of this feature:
...
+- `production`: The eMMC RPMB partition is used with the assumption that the RPMB key has already been programmed (enrolled). Designed for production environments where secure storage relies on a pre-provisioned RPMB key.
+
+With this, one could use an OS image configured in `factory` mode during a secure provisioning process, then deploy an OS image configured in `production` mode to devices in the field. However, be aware that alternative RPMB key provisioning methods exist, and the approach should align with your threat model.
+
+Another important aspect is how the key generation works. Because OP-TEE derives the RPMB authentication key from the SoC's HUK, if the RPMB key is provisioned while the device is in an "open" state (secure boot disabled), it becomes inaccessible once secure boot is enabled, because the HUK changes when the device transitions to a "closed" state.
```
rborn-tx, line 97: @sergioprado For clarification, for Toradex modules I imagine the procedure for provisioning a device in factory would involve (at least for NXP modules):

- Installing a version of the OS built with `TDX_OPTEE_FS_RPMB_MODE=factory` with Toradex Easy Installer.
- Booting that version (twice) to fuse and close the device.
- Booting that version again to write the private key to the RPMB partition. Btw, would this writing happen automatically?
- With the device closed, install a new version of the OS now built with `TDX_OPTEE_FS_RPMB_MODE=production`.

Does this seem correct? If so, then I'd say customers would not be able to perform the last step easily since it would require either a signed version of Toradex Easy Installer (which we don't provide yet) or provisioning the device to the platform and performing a normal (possibly offline) update. Not that this is a major issue since everything on Secure Boot is a work in progress :slightly_smiling_face:.

rborn-tx (commented):
This is a nice addition to the layer.

I have raised only a clarification question and a suggestion.

docs/README-optee.md
sergioprado: @rborn-tx Alright, let me break this down and clarify the process for RPMB key provisioning.

To write to the RPMB partition, an authentication key must first be enrolled (written to the partition). This is a one-time operation - once the key is programmed, it cannot be changed or erased. After enrollment, only devices with the correct key can perform authenticated read and write operations to the RPMB.

There are multiple ways to handle RPMB key provisioning. You could, for example:

- Program the key during eMMC manufacturing.
- Use a specialized firmware for that.
- Create a custom OS image designed to write the key.

The specific method doesn't matter as long as it fits your requirements. Ultimately, it's up to the user to choose an approach that works within their production environment.

Anyway, for secure storage to function effectively, OP-TEE needs access to the RPMB authentication key. It's impractical (and insecure) to hard-code this key in OP-TEE. Instead, OP-TEE can derive the key from the SoC's Hardware Unique Key (HUK), which is securely stored within the chip. When the `CFG_RPMB_WRITE_KEY` configuration is enabled, OP-TEE automatically derives and programs the key at startup.

A crucial point to understand is that you don't want to do this in production. OP-TEE itself does not have direct access to the eMMC driver and therefore cannot send commands to the RPMB partition. It delegates this task to the TEE supplicant running in the normal (REE) world. If this key-writing functionality is left enabled in a production environment, there is a risk that an attacker could exploit this "window of opportunity" to compromise the key by intercepting commands sent through the normal world.

To mitigate this risk, I introduced the concept of `factory` and `production` modes:

- Factory Mode: This mode allows OP-TEE to derive and program the RPMB key. It is intended to be used only during secure provisioning, such as in a factory setting.
- Production mode: In this mode, OP-TEE assumes the RPMB key has already been enrolled. The system can securely access the RPMB partition but cannot write or overwrite the key. This mode is designed for field devices to prevent key exposure.

However, this is just one possible solution. There are other approaches to RPMB key provisioning, which is why I emphasized in the documentation:

> Be aware that alternative RPMB key provisioning methods exist, and the approach should align with your threat model.

To summarize, the complete process would be:

- Fuse keys and close the device.
- Write the RPMB key (derived from the HUK) in a secure provisioning environment (e.g., using the factory image).
- Transition the device to production mode with key provisioning disabled.

Does this make sense? Please let me know if you have further questions.

Thanks for reviewing!

docs/README-optee.md
sergioprado: Done!

rborn-tx (approved):
@sergioprado I'm approving this PR since the questions I have are mainly clarifications that do not prevent merging it. I'll continue the clarification/discussions in the corresponding threads.

docs/README-optee.md
rborn-tx: @sergioprado Thanks for the detailed answer so far :slightly_smiling_face:.

Regarding this sentence in the documentation:

> factory: The eMMC RPMB partition is used, and OP-TEE is configured to program the RPMB key, derived from the SoC's Hardware Unique Key (HUK) and other device specific information.

When does this programming happen? Based on your last answer, it seems OP-TEE would not be able to do it alone during boot for example since it relies on operations provided by the OS (REE), right? Some possibilities that I imagine, all considering `TDX_OPTEE_FS_RPMB_MODE=factory`: (a) the programming happening when the "supplicant" software starts in the OS; or (b) in the first attempt to write to the RPMB by some application or (c) by running some special application. Anyway, maybe that's something to be mentioned in the documentation?

--
Apart from the question above, another thing I wanted to do is to translate the detailed information we're providing in the documentation of the present layer into steps to be followed by Toradex users, particularly for users of Torizon OS. Would that be possible? Or do we miss anything in our solution so users can easily leverage the feature being introduced here?

sergioprado: > When does this programming happen? Based on your last answer, it seems OP-TEE would not be able to do it alone during boot for example since it relies on operations provided by the OS (REE), right? Some possibilities that I imagine, all considering TDX_OPTEE_FS_RPMB_MODE=factory: (a) the programming happening when the "supplicant" software starts in the OS; or (b) in the first attempt to write to the RPMB by some application or (c) by running some special application. Anyway, maybe that's something to be mentioned in the documentation?

At boot time, when the OP-TEE supplicant daemon is started in REE, it sends some messages to OP-TEE OS to initialize the storage, and if CFG_RPMB_WRITE_KEY is enabled, the process to write the RPMB key is started.

> Apart from the question above, another thing I wanted to do is to translate the detailed information we're providing in the documentation of the present layer into steps to be followed by Toradex users, particularly for users of Torizon OS. Would that be possible? Or do we miss anything in our solution so users can easily leverage the feature being introduced here?

This layer just provide the infrastructure to facilitate enabling security features available in the hardware and/or software platform. It doesn't try to be opinionated - though it provide some sane defaults.

Having said that, the answer to your question is yes: Torizon OS users might use this layer as is and just build something nice on top of it. :-) 

But be aware that, though this layer tries to be generic for most use cases, it doesn't cover all use cases. For example, we don't enable OP-TEE tooling in U-Boot. If you need access to OP-TEE via U-Boot, then some extra work might be required. And patches are welcome. :-) 

Please let me know if you have any further questions.

PR #91, opened 2025-02-06
Ensure BSP DATA partition variables match with tdx-tezi-data-partition and tdx-encrypted classes.

MMG-DG:
**Environment:**
Image: tdx-reference-minimal-image 
Branch: Kirkstone

**Issue:**
When combining tdx-tezi-data-partition and tdx-encrypted classes in a build, the values defined for the `TEZI_DATA_MOUNTPOINT` do not match what was specified for `TDX_TEZI_DATA_PARTITION_MOUNTPOINT` or `TDX_ENC_STORAGE_MOUNTPOINT` respectivley.

**Solution:**
- Added addition to tdx-tezi-data-partition.bbclass to set the `TEZI_DATA_MOUNTPOINT` based on the value of `TDX_TEZI_DATA_PARTITION_MOUNTPOINT`.
- Added wrapper class to ensure correct flow-down of `TDX_ENC_STORAGE_MOUNTPOINT` to `TDX_TEZI_DATA_PARTITION_MOUNTPOINT` when encryption is explicitly wanted for the data partition.

### DIFF
--- a/classes/tdx-encrypted-data-partition.bbclass
+++ b/classes/tdx-encrypted-data-partition.bbclass
@@ -1,0 +1,4 @@
+inherit tdx-tezi-data-partition
+
+inherit tdx-encrypted
+TDX_TEZI_DATA_PARTITION_MOUNTPOINT = "${TDX_ENC_STORAGE_MOUNTPOINT}"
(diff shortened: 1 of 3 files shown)


rborn-tx:
Hi @MMG-DG. Wanted to confirm: are you expecting a review of this work? I see this relies on a variable `TEZI_DATA_MOUNTPOINT` that doesn't exist at the BSP level, right?

MMG-DG:
> Hi @MMG-DG. Wanted to confirm: are you expecting a review of this work? I see this relies on a variable `TEZI_DATA_MOUNTPOINT` that doesn't exist at the BSP level, right?

@rborn-tx - Indeed it would be useful to have review comments.
Although this does indeed set a variable value, the fact this doesn't currently exist in the BSP just means that setting the value will have no effect.

I have submitted a patch for the BSP which introduces this variable and the related changes to enable auto generation of a datafs image that takes the specified mountpoint from the rootfs and extracts it for the data partition as part of the tezi image generation.

I am happy to put this into "Draft" if preferred until any BSP changes are in place.

rborn-tx:
> I am happy to put this into "Draft" if preferred until any BSP changes are in place.

Yes, I'd suggest to put it into Draft state for now.

sergioprado:
Closing this due to a lack of activity in recent months.

PR #92, opened 2025-02-20
Add OP-TEE support for Verdin AM69

sergioprado:
This adds OP-TEE support for Verdin AM69, including RPMB as a secure storage device, fTPM and PKCS#11 trusted applications.


rborn-tx (approved):
LGTM

PR #93, opened 2025-02-26
u-boot-harden: add variable to enable/disable hardening debugging

rborn-tx:
Add new OE variable "TDX_UBOOT_HARDENING_ENABLE_DBG" to control whether the debugging of the hardening feature is enabled or not.

On the U-Boot side, the debugging has been available since the initial implementation of the hardening but in order to leverage it one would need to create a config fragment and apply it to U-Boot, which was cumbersome. With this commit we make the enabling simpler with the hope that users won't shoot their own feet by leaving this enabled in production.


sergioprado (approved):
Nice addition! Looks good to me.

jsrc27 (approved):

PR #94, opened 2025-02-27
u-boot-harden: simplify command whitelist configuration

rborn-tx:
Add new OE variables to allow the whitelist feature to be easily configured:

- TDX_SECBOOT_WL_ALLOW_OPEN_CATEG
- TDX_SECBOOT_WL_ALLOW_CLOSED_CATEG
- TDX_SECBOOT_WL_DENY_OPEN_CATEG
- TDX_SECBOOT_WL_DENY_CLOSED_CATEG
- TDX_SECBOOT_WL_NEEDED_CATEG

With this commit these variables will define the values that go inside the U-Boot control DTB which in turn control what commands are allowed or denied by the feature. Before, a user would need to patch or replace the control DTB fragment which was cumbersome.

The documentation is also updated to describe the variables and refer to a more detailed explanation at the Toradex website.


sergioprado (approved):
Nice addition!

LGTM.

jsrc27 (approved):

PR #98, opened 2025-03-13
Add explicit self-overwrite safeguard to "load" command plus fixes to built-in protection

rborn-tx:
This introduces two changes:

* Add a new protection to the `load` command to limit to what addresses a file can be loaded to; this is an additional safeguard to the built-in protections in U-Boot (which are buggy by the way) - the corresponding patch is applied only when the hardening is enabled.
* Fix the original built-in protection - the related patch is applied if secure-boot is enabled (whether the hardening is enabled or not); the fix is applied only to NXP upstream and TI downstream U-Boot since a different solution was employed by NXP on their downstream version of U-Boot.

For details see the individual commit messages.

Related-to: TOR-3771

### DIFF
--- a/recipes-bsp/u-boot/files/0001-lmb-prevent-alloc-overlap-with-reserved-regions-k3.patch
+++ b/recipes-bsp/u-boot/files/0001-lmb-prevent-alloc-overlap-with-reserved-regions-k3.patch
@@ -1,0 +1,84 @@
+From 3c86dfe96b06352231670bdbc47a06c0a8b01c40 Mon Sep 17 00:00:00 2001
+From: Rogerio Guerra Borin <rogerio.borin@toradex.com>
+Date: Wed, 12 Mar 2025 19:47:36 -0300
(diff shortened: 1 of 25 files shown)


sergioprado (approved):
Nice additions! LGTM.

Are there any plans to upstream "u-boot: fix built-in protection against self-overwriting"?

rborn-tx:
> Are there any plans to upstream "u-boot: fix built-in protection against self-overwriting"?

@sergioprado I intended to try at least. I mean, I believe the following patches are generic enough to be acceptable upstream:

- 0006-toradex-cmd-fuse-add-switch-for-quiet-operation-upst.patch
- 0007-toradex-cmd-fuse-add-readv-command-upstream.patch
- 0001-lmb-prevent-alloc-overlap-with-reserved-regions-upst.patch (related to the commit you asked)

But of course only after trying will I know if they agree with my opinion :grin:

jsrc27 (approved):

PR #108, opened 2025-04-21
Allow targeting HS-FS or HS-SE devices (TI / am62)

rborn-tx:
Originally we produced images targeting HS-SE (closed) devices which was focused on production. With this PR, we allow the selection between HS-SE or HS-FS.

To avoid confusion, some variables have been renamed which means users will need to update their configurations (i.e. we are introducing breaking changes). To provide a smooth-ish transition we show an error message and stop be build if we detect that the old variables are set. This will lead users to review their configurations.

See the commit messages for further details.

Related-to: TOR-3805


rborn-tx:
@sergioprado @jsrc27 After reviewing, please don't merge it yet. My idea is to synchronize this with the corresponding changes in Jenkins to avoid breaking the verdin-am62 builds.

jsrc27 (approved):
LGTM

classes/tdx-signed-k3-secboot.inc
```diff
@@ -7,5 +7,40 @@ TDX_K3_SECBOOT_TARGET_HSSE_DEVICE ?= "0"
 # Signing keys location.
 TDX_K3_SECBOOT_KEY_DIR ?= "${TOPDIR}/keys/ti"
 
+# Whether to show warnings about the generated image with regards to
+# variable TDX_K3_SECBOOT_TARGET_HSSE_DEVICE:
+TDX_K3_SUPPRESS_HSSE_WARNINGS ?= "0"
```
sergioprado, line 12: Great idea to introduce a variable for suppressing warnings. I was wondering if it might make sense to have a single variable that suppresses all warnings across the layers, rather than just this one. There are other places in the layers where we also display warning messages, so something like `TDX_META_TORADEX_SECURITY_SUPPRESS_WARNINGS` could be more consistent.

This way, we could apply it to all areas where warnings are shown.

What do you think?

sergioprado (approved):
I have just one suggestion for improvement. Other than that, it looks good to me.

classes/tdx-signed-k3-secboot.inc
rborn-tx: @sergioprado Thanks for your suggestions.

Thinking about it, my concern is that users could miss important warnings if we had such a global suppression variable. I mean, a user on a previous version of the layer could look at the warnings and decide to suppress them because they are not relevant to their use case but then later when they upgrade the layer to a new version having some important new warning they would not see it because of the previous setting. Having more granular control of the warnings avoids or at least minimizes the chances of this happening, don't you think?

sergioprado: Hmm, that makes total sense. Lets keep that way then.

rborn-tx:
I'm moving this to Draft state to avoid merging until the MR on Jenkins is reviewed.

PR #109, opened 2025-04-24
Fixes for AHAB support on iMX95

sergioprado:
Two additional changes were required to enable boot container signature via AHAB for iMX95:

- Disabled SGK for iMX9-based SoCs because it is currently not supported in the current firmware for SoCs using the EdgeLock Secure Enclave (ELE), which includes i.MX9x SoCs.
- On iMX95, the u-boot-atf-container.img is generated separately and later appended to flash.bin. To ensure proper signing, it must be signed before generating and signing the final boot container image.

With this, signing the boot container works as expected:

```
u-boot=> ahab_status
Lifecycle: 0x00000010, OEM Open

        No Events Found!
```

The changes were tested on the iMX95 Verdin EVK, as well as on the Apalis iMX8 to to make sure no regressions were introduced.


dynamic-layers/freescale/recipes-bsp/imx-mkimage/files/mx8_sign.sh
```diff
@@ -110,18 +110,19 @@ generate_csf_ahab() {
...
     sed -i "s|@@CST_KIDX@@|${kidx}|g" "${image_csf}"
 
-    if [ "${TDX_IMX_HAB_CST_SRK_CERT##*_ca_}" = "crt.pem" ]; then
+    if [ "${TDX_IMX_HAB_CST_SGK_SUPP}" = "0" ] || [ "${TDX_IMX_HAB_CST_SRK_CERT##*_usr_}" = "crt.pem" ]; then
+        # SGK not supported or file name ends with "_usr_crt.pem": an SGK is not expected.
+        # Remove SGK block from CSF.
+        echo "Inferred that CA flag was not set; signing with SRK only."
+        sed -i "/#+START_SGK_BLOCK/,/#+END_SGK_BLOCK/d" "${image_csf}"
+
+    elif [ "${TDX_IMX_HAB_CST_SRK_CERT##*_ca_}" = "crt.pem" ]; then
```
rborn-tx, line 119: Since the ROM code does not support the use of the SGKs, I'd tend to say that users trying to use them are basically doing something wrong/unsupported, probably because they are unaware of that lack of support. Considering that, shouldn't we throw an error instead? Otherwise, we'd simply ignore the SGKs that the user believes are in use.

By throwing an error we'd force users to generate the keys indicating that the CA is not in use which should also avoid the case where they could try to reuse keys they previously created for other AHAB devices (possibly with unsupported algorithm which we know would cause issues). What do you think?

rborn-tx (commented):
I've left a question/suggestion for evaluation mainly related to UX.

dynamic-layers/freescale/recipes-bsp/imx-mkimage/files/mx8_sign.sh
jsrc27: Further related to this point. If for some reason by purpose/accident a user overrides `TDX_IMX_HAB_CST_SGK_SUPP` to 1 even when building for an i.MX9* I think that should trigger some kind of failure/error to notify the user as well.

sergioprado: That’s an interesting topic, and I’ve given it some thought.

First, it’s important to note that AHAB support on the iMX95 is scarcely documented in NXP's resources (e.g., the CST manual or AHAB application note). I found only one post on the NXP community website discussing secure boot support for the iMX95. When asked about this, NXP tech support just answered, "The i.MX95 supports secure boot, but as the device is still in preproduction, please wait for the official release." 😲

https://community.nxp.com/t5/i-MX-Processors/IMX95-SECURE-BOOT-AHAB-Support/m-p/2080439

Anyway, my conclusions are based on the fact that iMX95 uses the EdgeLock Secure Enclave (ELE), which is a feature also present in other SoCs like the iMX93 and i.MX8ULP. And those SoCs are mentioned in the docs.

Now, in the latest release of the AN12312 document ("Secure Boot on AHAB Supported Devices - Rev. 1 — 14 August 2023"), page 11, it states:

> Note: For i.MX8ULP and i.MX93, only SRK is supported in the current released firmware.

Based on this, here are my assumptions:

1. Since the document specifically mentions i.MX8ULP and i.MX93, I’m assuming the same applies to iMX95, as they all share the EdgeLock Secure Enclave.
2. The mention of `firmware` suggests that the limitation is within the EdgeLock Secure Enclave (ELE) firmware, not the ROM, so it could be updated/changed in the future.
3. The use of the word `current` implies that this limitation may change in the future.

Given this, I decided to create a variable to enable or disable SGK usage, making the solution more future-proof. However, I understand your concern that users might mistakenly believe SGK is being used.

Possible solutions:

1. Add a warning during the build process.
2. Include a note in the documentation.
3. Avoid using a keyset that contains SGK (as you suggested), and fail the build.
4. Something else?

What do you guys think?

tkfu: Well, first, I think we should definitely put a note in the documentation. (Also, I maintain that for 99.99% of users, using SGKs is pointless, but there's no need to rehash that argument here.)

On the merits: the ELE firmware is in the part of the boot container that's signed by the SRK only, right? And then the SPL validates TF-A and u-boot proper, probably calling into the ELE. That's what I understand from the [NXP doc](https://github.com/nxp-imx/uboot-imx/blob/lf_v2023.04/doc/imx/ahab/guides/mx8ulp_9x_secure_boot.txt#L111), at least. So I think Sergio's points are strong, and it seems likely that they implement the support eventually.

That said, I do agree with Rogerio's point that silently discarding the SGK in this case could be subverting the user's expressed intent.

I think it's sufficient if we add a warning, though, rather than failing the build. People might want to just try out building an image with their dev signing keys rather than having to generate new ones. Plus, it is still just an EVK, after all, so I think it makes a certain amount of sense to be more flexible.

I suggest something like "WARNING: AHAB (Secure Boot) signing keys have the CA flag enabled, but the build target $MACHINE does not support the use of subordinate signing keys. Bootloader will be signed using SRK directly."

rborn-tx: Ah, the joys of living on the cutting edge :laughing: 

The use a variable to indicate if the support for the SGK exists makes sense to me.

So, based on what you said it is possible that at some point we'd need to change that variable to indicate that the support exists. If the ELE firmware is flashed into the device, I can imagine the following situation:

A user is building images for devices without the SGK support with current default value of `TDX_IMX_HAB_CST_SGK_SUPP` for the i.MX95 (`0`) and this works because their devices don't have that support. At some point we change that default to `1` because Toradex started shipping devices with a new firmware. If that user upgrades the layer, the newly built images will not longer work on their devices.

Is this a big issue? Probably not because the user should be able to see that happening while the devices are in the factory and then they could force the setting of the variable back to `0` and fix the situation for them.

Anyway, for me it seems we cannot transparently solve the situation. Thus, the best we can do is to ensure users are aware of it. For this, I believe having a warning to say that the SGK is being ignored (i.e. the user is trying to use the SGK but the support does not exist) and adding a note to the documentation should be enough. As I said, since the situation is fixable (without losing a device) I don't think we need the fatal error.

Makes sense?

PS.: It looks like I was writing this while the answer from Jon wasn't still present on the page but it seems we basically agree.

tkfu: > A user is building images for devices without the SGK support with current default value of TDX_IMX_HAB_CST_SGK_SUPP for the i.MX95 (0) and this works because their devices don't have that support. At some point we change that default to 1 because Toradex started shipping devices with a new firmware. If that user upgrades the layer, the newly built images will not longer work on their devices.

I believe the way these edgelock-in-the-soc designs work is that they have a two-layer firmware: there's some core functionality that's in mask-programmed ROM and can never be altered, and then there's a firmware layer that can be used to implement more complex functionality. As far as I can tell, that second firmware layer is included in the boot container, in the portion of the container that must be signed by the SRK, and then it's loaded from there on each boot (see the diagram at the link in my previous post). So the situation you outline here wouldn't happen, because the ELE firmware isn't flashed persistently. Or at least, that'd be the case if my understanding is correct, but it's hard to say with 100% confidence because the documentation isn't the greatest.

sergioprado: Thank you all for the feedback!

I added a warning message at build time and a note in the documentation.

The pull request is updated and ready for another review!

docs/README-secure-boot-imx.md
```diff
@@ -37,6 +37,7 @@ The complete list of variables can be found in the `imx-hab.bbclass` file.
 
 - On devices based on the NXP i.MX6 and i.MX7 SoCs, `TDX_IMX_HAB_CST_SRK_CA` must be set to `1` and the SRK certificates must be generated with the CA flag set. Build errors will happen if the variable is set to `0` because the signing logic currently unconditionally tries to use the CSF and IMG certificates which are not generated by CST when the CA flag is not set.
 - Starting from version 4.0.0, the NXP CST tool may not work as expected on older Linux distributions (e.g., Ubuntu 20.04). If you are using NXP CST 4.0.0 or later, it is recommended to use a more recent Linux distribution (e.g., Ubuntu 24.04) to ensure compatibility and avoid potential issues.
+- As of April 2025, SGK is not supported in the current firmware for SoCs using the EdgeLock Secure Enclave (ELE), which includes iMX8ULP and iMX9x. Consequently, when building for machines based on these SoCs, the use of subordinate signing keys (SGK) will be disabled, and the boot container will be signed directly with the signing root keys (SRK).
```
rborn-tx: signing root keys -> Super Root Keys

dynamic-layers/freescale/recipes-bsp/imx-mkimage/imx-boot-hab.inc
```diff
@@ -77,6 +77,13 @@ sign_common() {
         bberror "Could not find SRK fuse file '${TDX_IMX_HAB_CST_SRK_FUSE}'."
     fi
 
+    local srk_cert=${TDX_IMX_HAB_CST_SRK_CERT}
+    if [ "${srk_cert##*_ca_}" = "crt.pem" ] && [ "${TDX_IMX_HAB_CST_SGK_SUPP}" = "0" ]; then
+        bbwarn "AHAB (Secure Boot) signing keys seems to have the CA flag enabled, but" \
```
rborn-tx: seems -> seem

rborn-tx (commented):
Found some typos. Other than that, LGTM.

docs/README-secure-boot-imx.md
sergioprado: Fixed, thanks!

dynamic-layers/freescale/recipes-bsp/imx-mkimage/imx-boot-hab.inc
sergioprado: Fixed, thanks!

sergioprado:
@rborn-tx fixed the typos and updated the pull request.

rborn-tx (approved):

jsrc27 (approved):

PR #110, opened 2025-05-01
Secure boot support on iMX95 Verdin EVK

sergioprado:
This pull request adds support for full chain-of-trust (aka secure boot) on iMX95 Verdin EVK.

All tests were done using NXP BSP (machine definition in `meta-imx-bsp`).

Also tested this on Verdin iMX8MP to make sure there were no visible regressions.


docs/README-secure-boot-imx.md
```diff
@@ -8,7 +8,7 @@ To support bootloader signature checking, such a feature needs to be available i
 
 On iMX6, iMX7 and iMX8M, this feature is available and it is called HAB (High Assurance Boot).
 
-On iMX8 and iMX8X, this feature is available and it is called AHAB (Advanced High Assurance Boot).
+On iMX8, iMX8X and iMX9X, this feature is available and it is called AHAB (Advanced High Assurance Boot).
```
jsrc27: Minor suggestion here.

I assume the idea was to use 'iMX9X" as a wildcard to refer to the entire iMX9 SoC family. But I got confused at first cause I thought this referred to a specific SoC named the iMX9X. I think what made it confusing was right before the iMX8X gets referred to which is a specific SoC.

jsrc27 (approved):
LGTM. Just left a minor non-critical suggestion.

docs/README-secure-boot-imx.md
sergioprado: You are right! It is indeed confusing. I changed to iMX95 and updated the pull request. Thanks for poiting that out!

rborn-tx (approved):
Nice refactoring of the include files!

LGTM

PR #111, opened 2025-05-05
tdx-signed: fix HSSE/HSFS selection

rborn-tx:
Commit 252bf2e (classes: improve organization of include files) caused a regression on the HSSE/HSFS selection because the include file defining the FIRMWARE_BINARY names was made independent of the variable `TDX_K3_SECBOOT_TARGET_HSSE_DEVICE` (i.e. the binary names were set for HSSE regardless of the image targeting an HSSE device or not). Here we restore the previous/desired behavior.

Related-to: TOR-3835

### DIFF
--- a/classes/include/signed/machine/aquila-am69.inc
+++ b/classes/include/signed/machine/aquila-am69.inc
@@ -1,2 +1,2 @@
 # use tiboot3.bin secure boot variant in Tezi
-FIRMWARE_BINARY[0088] = "tiboot3-am69-hs-aquila.bin"
+FIRMWARE_BINARY[0088] := "tiboot3-am69-${SYSFW_SUFFIX}-aquila.bin"
--- a/classes/include/signed/machine/verdin-am62.inc
+++ b/classes/include/signed/machine/verdin-am62.inc
@@ -1,7 +1,7 @@
 # use tiboot3.bin secure boot variant in Tezi
-FIRMWARE_BINARY[0071] = "tiboot3-am62x-hs-verdin.bin"
-FIRMWARE_BINARY[0072] = "tiboot3-am62x-hs-verdin.bin"
-FIRMWARE_BINARY[0073] = "tiboot3-am62x-hs-verdin.bin"
-FIRMWARE_BINARY[0074] = "tiboot3-am62x-hs-verdin.bin"
-FIRMWARE_BINARY[0075] = "tiboot3-am62x-hs-verdin.bin"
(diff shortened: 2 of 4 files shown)


jsrc27 (approved):

sergioprado:
Hmmm, that is really unfortunate. Anyway, for better organization and improved maintenance, I would really like to keep the current structure of include files, keeping all machine-related metadata separated in the `machine` directory.

I have a different approach I would like to explore.

What if we add the following configuration to `tdx-signed-k3-secboot.inc`, just before including `tdx-signed-k3-secboot-tiboot3.inc`:

```
# Use non-secure variant of tiboot3.bin by default
SYSFW_SUFFIX ?= "hs-fs"
```

And then use `SYSFW_SUFFIX` in the `FIRMWARE_BINARY` variables:

```
FIRMWARE_BINARY[0071] = "tiboot3-am62x-${SYSFW_SUFFIX}-verdin.bin"
```

What do you think?

rborn-tx:
> Hmmm, that is really unfortunate. Anyway, for better organization and improved maintenance, I would really like to keep the current structure of include files, keeping all machine-related metadata separated in the `machine` directory.
> 
> I have a different approach I would like to explore.
> 
> What if we add the following configuration to `tdx-signed-k3-secboot.inc`, just before including `tdx-signed-k3-secboot-tiboot3.inc`:
> 
> ```
> # Use non-secure variant of tiboot3.bin by default
> SYSFW_SUFFIX ?= "hs-fs"
> ```
> 
> And then use `SYSFW_SUFFIX` in the `FIRMWARE_BINARY` variables:
> 
> ```
> FIRMWARE_BINARY[0071] = "tiboot3-am62x-${SYSFW_SUFFIX}-verdin.bin"
> ```
> 
> What do you think?

That should work fine as well. I'll give it a try. Thanks for the idea!

sergioprado (approved):
Loved the logic to remove the extra include file! :+1:

rborn-tx:
@sergioprado This is ready for review after making the changes to follow your suggestions.

One note: I had to replace the assignment operators on the `FIRMWARE_BINARY` varflags by `:=` otherwise the references to the variable `SYSFW_SUFFIX` did not get expanded. I think the root reason for this is that the code that reads the varflags (in `meta-toradex-bsp-common`) does not invoke `d.expand()` on the varflag value...

sergioprado:
@rborn-tx LGTM.

Before merging this, let's see if @jsrc27 has any comments on the updated PR.

PR #113, opened 2025-05-08
u-boot-harden: fixes for devices having slightly less than 1GB RAM

rborn-tx:
Change the logic responsible for defining the offset limit of the "load" command (which is part of the self-write protection). The old logic had two hard-coded limits: one that was selected if the device had 1GB of memory (or more) and another for when it had less than 1GB, which was supposed to cover the devices with 512MB of RAM. However, some devices with 1GB of RAM reserve some small amount of memory for particular purposes which causes the old logic to fail by selecting the limit as if the device had 512MB of RAM. This was found to prevent the boot of those devices (particularly some PID4s of the colibri-imx8x). With the new logic the limit is calculated from the available memory as the minimum between the available RAM size and 1GB, minus 64MB.

Related-to: TOR-3830


sergioprado (approved):
LGTM.

jsrc27 (approved):
LGTM

PR #114, opened 2025-05-14
Allow signing bootloader artifacts with SRK2..4 on HABv4

rborn-tx:
This impacts the NXP machines based on i.MX6/6ULL/7 only.

- Add support for signing the bootloader binary with an SRK other than the first one (SRK1).
- Add support for signing with the SRKs only, i.e. without the subordinate keys (corresponding to the NOCAK case offered by NXP) - not tested at the moment but code matches exactly what was done for the i.MX8M where the feature was once tested.

For details, please refer to the commit messages.

The following cases have been tested:

| Machine              |            | Build | Check CSF | hab_status |
| -------------------- | ---------- | ----- | --------- | ---------- |
| colibri-imx6         | CAK/SRK1   | OK    | OK        | OK         |
| colibri-imx6ull-emmc | CAK/SRK1   | OK    | OK        | OK         |
| colibri-imx7-emmc    | CAK/SRK1   | OK    | OK        | OK         |
|                      | CAK/SRK3   | OK    | OK        | OK         |
|                      | NOCAK/SRK1 | OK    | OK        | -          |

### DIFF
--- a/recipes-bsp/u-boot/u-boot-hab.inc
+++ b/recipes-bsp/u-boot/u-boot-hab.inc
@@ -47,13 +47,6 @@ imx6_imx7_sign_habv4() {
         bbfatal "Could not find CST binary at ${TDX_IMX_HAB_CST_BIN}."
     fi
 
-    for f in "${TDX_IMX_HAB_CST_SRK}" "${TDX_IMX_HAB_CST_CSF_CERT}" \
-             "${TDX_IMX_HAB_CST_IMG_CERT}" "${TDX_IMX_HAB_CST_SRK_FUSE}"; do
-        if [ ! -e "${f}" ]; then
-            bbfatal "Could not find file '${f}' (required for HAB)."
-        fi
-    done
-
     bbdebug 1 "Generating CSF for U-Boot"
     # Generate CSF file:
     TDX_IMX_HAB_CST_SRK="${TDX_IMX_HAB_CST_SRK}" \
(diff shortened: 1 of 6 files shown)


jsrc27:
LGTM no major comments.

One question though are we going to update the in-code comments regarding the `TDX_IMX_HAB_CST_KEY_INDEX` variable here: https://github.com/toradex/meta-toradex-security/blob/scarthgap-7.x.y/classes/imx-hab.bbclass#L39

Including documenting this variable in the README as I assume that's how we intend for users to choose which SRK to use.

Or will the documentation be updated in a subsequent PR?

recipes-bsp/u-boot/u-boot-hab.inc
```diff
@@ -47,16 +47,25 @@ imx6_imx7_sign_habv4() {
         bbfatal "Could not find CST binary at ${TDX_IMX_HAB_CST_BIN}."
     fi
 
-    for f in "${TDX_IMX_HAB_CST_SRK}" "${TDX_IMX_HAB_CST_CSF_CERT}" \
-             "${TDX_IMX_HAB_CST_IMG_CERT}" "${TDX_IMX_HAB_CST_SRK_FUSE}"; do
+    for f in "${TDX_IMX_HAB_CST_SRK}" "${TDX_IMX_HAB_CST_SRK_CERT}" \
+             "${TDX_IMX_HAB_CST_SRK_FUSE}"; do
```
sergioprado: Since you are already checking for the existence of the files in the signing script (function `validate_environ`), makes sense to remove the checks here?

rborn-tx:
> One question though are we going to update the in-code comments regarding the `TDX_IMX_HAB_CST_KEY_INDEX` variable here: https://github.com/toradex/meta-toradex-security/blob/scarthgap-7.x.y/classes/imx-hab.bbclass#L39
> 
> Including documenting this variable in the README as I assume that's how we intend for users to choose which SRK to use.
> 
> Or will the documentation be updated in a subsequent PR?

Yes, my idea is to update the documentation after the features are tested on AHAB as well.

I say tested because in the past, when I touched the signing code for AHAB, I kind of implemented both of the features being introduced here for the i.MX6/6ULL/7 but I never tested them in detail. So, my next step is to run tests on AHAB, apply fixes (if necessary) and update the docs and in-code comments.

recipes-bsp/u-boot/u-boot-hab.inc
rborn-tx: Well, in theory we could remove the checks here and in fact I thought of doing it a couple of times but I didn't simply because the errors wouldn't stand out as much as they do today. That's the only downside I see. On the other hand, I see that maintenance would become simpler since we wouldn't have to keep these checks in sync with the signing script.

Thus, considering the above, we could decide to get rid of these checks and only keep the checks inside the script. Opinions? @sergioprado @jsrc27

sergioprado: Yeah, I think we don't need the double check. So we can remove the checks in the script or in the recipe. Up to you deciding where to keep them. :-)

jsrc27: I personally don't have any strong opinions here either way.

jsrc27 (approved):

recipes-bsp/u-boot/u-boot-hab.inc
rborn-tx: Done

sergioprado (approved):
LGTM.

PR #115, opened 2025-05-16
Enable encryption support for iMX95 Verdin EVK

sergioprado:
Since the i.MX95 lacks the CAAM hardware module, encryption key management is implemented using the EdgeLock Enclave (ELE) in combination with OP-TEE. This setup allows secure generation, storage, and usage of encryption keys through the kernel’s trusted key infrastructure.

To support this, here we enable the support for using a Trusted Execution Environment (TEE), such as OP-TEE, as the trust anchor for key management via the kernel keyring service.

Additionally, a few changes are required in the kernel, OP-TEE, and the encryption handler script to ensure proper functionality with the NXP BSP on the i.MX95 Verdin EVK.

Build and runtime tested on i.MX95 Verdin EVK and Verdin iMX8MP.


docs/README-encryption.md
```diff
@@ -33,7 +33,7 @@ Trusted Keys make it possible to create and manage variable-length symmetric key
 
 Trusted Keys require the availability of a Trust Source for greater security. Different Trust Sources are supported, including CAAM (Cryptographic Acceleration and Assurance Module), TPM (Trusted Platform Module) and TEE (Trusted Execution Environment).
 
-This layer supports using CAAM and TPM as a source for managing the encryption key. CAAM is available on NXP iMX-based SoMs and TPM availability might depend on the selected SoM and carrier board.
+This layer supports using CAAM, TPM or TEE as a source for managing the encryption key. CAAM is available on NXP iMX-based SoMs and TPM availability might depend on the selected SoM and carrier board.
```
tkfu:
```suggestion
This layer supports using CAAM, TPM or TEE as a source for managing the encryption key. CAAM is available on NXP iMX-based SoMs (except for iMX95 and iMX6ULL) and TPM availability might depend on the selected SoM and carrier board.
```

docs/README-encryption.md
```diff
@@ -67,7 +67,7 @@ A few additional variables are available to customize the behavior of the data-a
...
-| `TDX_ENC_KEY_BACKEND` | Backend used to manage the encryption key. Allowed values: `caam`, `tpm` or `cleartext`. If configured with `caam`, it will use Trusted Keys backed by the CAAM device (available on NXP iMX-based SoMs). If configured with `tpm`, it will use Trusted Keys backed by a TPM device (availability depends on the hardware). If configured with `cleartext`, the encryption key will be stored in clear text in the file system (use `cleartext` only for testing purposes!) | `caam` on iMX based SoMs, empty otherwise |
+| `TDX_ENC_KEY_BACKEND` | Backend used to manage the encryption key. Allowed values: `caam`, `tpm`, `tee` or `cleartext`. If configured with `caam`, it will use Trusted Keys backed by the CAAM device (available on NXP iMX-based SoMs). If configured with `tpm`, it will use Trusted Keys backed by a TPM device (availability depends on the hardware). If configured with `tee`, it will use Trusted Keys backed by a Trusted Execution Environment. If configured with `cleartext`, the encryption key will be stored in clear text in the file system (use `cleartext` only for testing purposes!) | `caam` on iMX based SoMs, empty otherwise |
```
tkfu, line 71: But `tee` will only work if OP-TEE has the appropriate TAs running, right? I see that `CFG_HWRNG_PTA` and `CFG_IMX_TRUSTED_ARM_CE` are enabled only for the imx95 in `optee-tdx-imx95-19x19-verdin.inc`; wouldn't it make more sense to conditionally enable those options when the user has selected `tee` for `TDX_ENC_KEY_BACKEND`?

tkfu: Or alternatively, if this isn't sufficiently generic and can't be enabled on OP-TEE on other hardware and thus we can only make it available for the 95 (and any future NXP SoMs that use the edgelock enclave instead of the CAAM), maybe we could just call the option `edgelock` instead of `tee`?

docs/README-encryption.md
sergioprado: Thanks for the suggestion. I didn't add anything about iMX95 because I still plan to add another commit to this PR to document the iMX95 support (and meybe some generic info about ELE).

docs/README-encryption.md
sergioprado: > I see that CFG_HWRNG_PTA and CFG_IMX_TRUSTED_ARM_CE are enabled only for the imx95 in optee-tdx-imx95-19x19-verdin.inc; wouldn't it make more sense to conditionally enable those options when the user has selected tee for TDX_ENC_KEY_BACKEND?

Yeah, that might make sense. I didn't do this yet because I want to test this on a non ELE-based device.

> Or alternatively, if this isn't sufficiently generic and can't be enabled on OP-TEE on other hardware and thus we can only make it available for the 95 (and any future NXP SoMs that use the edgelock enclave instead of the CAAM), maybe we could just call the option edgelock instead of tee?

Here we are enabling the TEE backend for trusted keys (see https://www.kernel.org/doc/html/latest/security/keys/trusted-encrypted.html). On OP-TEE OS from NXP, the ELE driver is used for sealing/unsealing the encryption key, but still, the trusted source for the encryption key is TEE, so I think it makes more to keep it generic and use `tee`.

docs/README-encryption.md
sergioprado: Done!

docs/README-encryption.md
sergioprado: After some testing, I confirmed that `CFG_HWRNG_PTA` and other options could be moved to a generic config file for all OP-TEE builds. Done in the lasted PR.

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
```diff
@@ -58,4 +58,9 @@ do_install() {
         install -m 0644 ${WORKDIR}/99-tpm.rules ${D}${sysconfdir}/udev/rules.d/99-tpm.rules
         sed -i '/^After=/a Requires=dev-tpm0.device\nAfter=dev-tpm0.device' ${D}${systemd_system_unitdir}/tdx-enc-handler.service
     fi
+
+    if [ ${TDX_ENC_KEY_BACKEND} = "tee" ]; then
+        sed -i '/^Before=/d' ${D}${systemd_system_unitdir}/tdx-enc-handler.service
```
rborn-tx: Isn't this operation a bit brittle? I mean, today we want to remove the `Before=` line because the only dependency is on `local-fs.target`, right? But if we added other dependencies then we would remove all of them which could go unnoticed.

To avoid this, maybe we could first confirm if the line matches exactly `Before=local-fs.target` and, if not, throw an error saying this patching function needs a review. Does that make sense?

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
```diff
@@ -58,4 +58,9 @@ do_install() {
...
+        sed -i '/^After=/c\Requires=tee-supplicant.service\nAfter=tee-supplicant.service' ${D}${systemd_system_unitdir}/tdx-enc-handler.service
```
rborn-tx: If my previous comment make sense, then something similar would apply here.

rborn-tx (commented):
Nice!
I've added a suggestion but overall it looks good to me.

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
sergioprado: Yes, that’s a valid point. Do you think it would be better to have three variants of the service file? The downside is that we’d need to remember to update all of them when necessary, but this approach would allow us to eliminate the patching logic in the recipe.

sergioprado: I don't forsee future changes in this file, despite updates in its dependencies, which we would need do to it anyway in the recipe if we decide to keep the patching logic.

rborn-tx: Hmm, duplicating the service file would probably be worse for maintenance. The simplest solution I see is just adding a reminder for us to review the patching logic if those particular lines change, e.g.:

```
if ! grep -q '^Before=local-fs.target$' tdx-enc-handler.service; then
    bbfatal 'PLEASE REVIEW THIS LOGIC'
fi
if ! grep -q '^After=systemd-remount-fs.service$' tdx-enc-handler.service; then
    bbfatal 'PLEASE REVIEW THIS LOGIC'
fi

sed -i ...
```

recipes-kernel/linux/files/imx95-19x19-verdin/tee.cfg
```diff
@@ -0,0 +1,5 @@
+CONFIG_TRUSTED_KEYS_TEE=y
+# CONFIG_TRUSTED_KEYS_CAAM is not set
+# CONFIG_TRUSTED_KEYS_TPM is not set
+CONFIG_IMX_SEC_ENCLAVE=n
```
jsrc27, line 4: Just to clarify, since I'm not too familiar with the i.MX95 ELE.

Do we disable it here in the kernel config, because OPTEE uses the ELE for the trusted keys? Or is there another reason we disable it?

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
sergioprado: Alright. I had another idea. Instead of patching, what if we move all the logic to setup the service dependencies to the install function?

We would remove `Before=`, `After=` and `Requires=` entries from the service file and replace them with a string that would be substitute later in the `install` function:

```
[Unit]
Description=Encryption handler for Toradex modules
DefaultDependencies=no
@@DEPENDENCIES@@
RequiresMountsFor=/tmp /var
```

Then we would add the logic in the `install` function:

```
if [ ${TDX_ENC_KEY_BACKEND} = "tpm" ]; then
    # add dependencies for TPM backend
fi

else if [ ${TDX_ENC_KEY_BACKEND} = "tee" ]; then
    # add dependencies for TEE backend
fi
    
else
    # add default dependencies
fi
```

recipes-kernel/linux/files/imx95-19x19-verdin/tee.cfg
sergioprado: > Just to clarify, since I'm not too familiar with the i.MX95 ELE. Do we disable it here in the kernel config, because OPTEE uses the ELE for the trusted keys? Or is there another reason we disable it?

Yes, exactly. We need to disable the driver in the kernel to allow OP-TEE to use it. This is documented in the i.MX Linux User's Guide (Rev. LF6.12.3_1.0.0, 21 April 2025) on page 129.

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
rborn-tx: Yeah, that's definitely a solution.

sergioprado: Done. Since the refactoring touched code unrelated to the introduced change, I added a new commit to this PR.

recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
sergioprado: Done.

jsrc27 (approved):

rborn-tx (approved):

PR #116, opened 2025-05-16
Allow signing bootloader artifacts with SRK2..4 on AHAB

rborn-tx:
The feature to allow signing with an SRK index other than 1 was actually already working with AHAB (from a previous implementation), but it had never been tested. However, here I took the opportunity to sync the signing script with the one for the i.MX8M which brought a few improvements to AHAB as well.

In summary this PR is adding a few non-functional improvements to the signing script and updating the documentation regarding the new variable `TDX_IMX_HAB_CST_SRK_INDEX` (the one allowing the selection of the SRK index). Btw, this variable was renamed from the previous implementation (where the variable wasn't documented yet). Anyways, I added a check to throw a fatal error in case the old variable is set.

Please see the individual commit messages for details.

This has been tested with the following configurations:

| Machine       |            | Build | Check CSF | ahab_status |
|---------------|------------|-------|-----------|-------------|
| colibri-imx8x | CAK/SRK1   | OK    | OK        | OK          |
|               | CAK/SRK3   | OK    | OK        | OK          |
|               | NOCAK/SRK2 | OK    | OK        | OK          |

### DIFF
--- a/classes/imx-hab.bbclass
+++ b/classes/imx-hab.bbclass
@@ -5,9 +5,10 @@ TDX_IMX_HAB_CST_DIR ?= "${TOPDIR}/keys/cst"
 TDX_IMX_HAB_CST_BIN ?= "${TDX_IMX_HAB_CST_DIR}/linux64/bin/cst"
 
 #
-# Main variables that control the automatic certificate names generation; users
-# will likely want to change these; they should match the answers/parameters
-# given to the CST tool when generating the keys and certificates.
+# Variables that control the automatic certificate names generation; those names
+# are used when signing with the CST tool. Users will likely want to change
+# these; they should match the answers/parameters given to the CST tool when
+# generating the keys and certificates.
 #
 # Explanation:
 #
@@ -21,29 +22,22 @@ TDX_IMX_HAB_CST_BIN ?= "${TDX_IMX_HAB_CST_DIR}/linux64/bin/cst"
 #   named "SRK1_sha256_secp384r1_v3_ca_crt.pem" (found in the certificates
 #   directory) the present variable would be set to "secp384r1".
 #
+# - TDX_IMX_HAB_CST_KEY_EXP: Key exponent for RSA keys (only).
+#
 # - TDX_IMX_HAB_CST_DIG_ALGO: Digest algorithm as entered into the CST tool.
 #
 # - TDX_IMX_HAB_CST_SRK_CA: Whether or not the SRK certificates have the CA flag
 #   set as entered into the CST tool; allowed values are "0" or "1"
 #
+# - TDX_IMX_HAB_CST_KEY_INDEX: Zero-based index of the SRK to be used within the
+#   SRK table.
+#
 TDX_IMX_HAB_CST_CRYPTO    ?= "rsa"
 TDX_IMX_HAB_CST_KEY_SIZE  ?= "2048"
+TDX_IMX_HAB_CST_KEY_EXP   ?= "65537"
 TDX_IMX_HAB_CST_DIG_ALGO  ?= "sha256"
 TDX_IMX_HAB_CST_SRK_CA    ?= "1"
-
-#
-# Secondary parameters for automaticaly generating certificate file names:
-#
-# Explanation:
-#
-# - TDX_IMX_HAB_CST_KEY_INDEX: Zero-based index of the SRK to be used within the
-#   SRK table. NOTE: This is not fully handled yet so it should be set to 0.
-# - TDX_IMX_HAB_CST_KEY_EXP: Key exponent for RSA keys (only).
-#
-
-# TODO: Handle the key index when generating the CSF files for HABv4 and AHAB.
 TDX_IMX_HAB_CST_KEY_INDEX ?= "0"
-TDX_IMX_HAB_CST_KEY_EXP   ?= "65537"
 
 #
 # Helper functions
--- a/classes/imx-hab.bbclass
+++ b/classes/imx-hab.bbclass
@@ -45,6 +45,13 @@ TDX_IMX_HAB_CST_SRK_INDEX ?= "1"
 #
 def make_srk_cert_name(d, basedir):
     """Generate certificate name related to a Super Root Key"""
+
+    # Check use of old variable:
+    if d.getVar('TDX_IMX_HAB_CST_KEY_INDEX'):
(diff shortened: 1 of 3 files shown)


classes/imx-hab.bbclass
```diff
@@ -72,12 +72,12 @@ def make_sub_cert_name(d, prefix, basedir):
...
+    kidx = int(d.getVar("TDX_IMX_HAB_CST_SRK_INDEX"))
     dalgo = d.getVar("TDX_IMX_HAB_CST_DIG_ALGO")
     ksize = d.getVar("TDX_IMX_HAB_CST_KEY_SIZE")
     caflg = int(d.getVar("TDX_IMX_HAB_CST_SRK_CA"))
     if kidx < 0 or kidx > 3:
-        bb.fatal("TDX_IMX_HAB_CST_KEY_INDEX must be in the range [0,3]")
+        bb.fatal("TDX_IMX_HAB_CST_SRK_INDEX must be in the range [0,3]")
     if caflg == 0:
         # Subordinate keys/certs only exist when the SRK cert has the CA flag set.
         pass
```
sergioprado, line 91: @rborn-tx Maybe I am missing something, but where in this commit you are a showing a fatal message in case the old variable is set?

sergioprado: Hmm, just saw that you are doing this in a separate commit. You probably need to update this commit's message?

rborn-tx: Nice catch! Updated.

docs/README-secure-boot-imx.md
```diff
@@ -28,8 +28,10 @@ After that, configure the various variables listed below to match your choices;
...
 | `TDX_IMX_HAB_CST_CRYPTO` | Type of cryptographic keys in use; allowed values: `rsa` or `ecdsa`. This should be set to `ecdsa` if (and only if) you selected "Elliptic Curve Cryptography" when generating the keys/certificates with the CST tool. | `rsa` |
 | `TDX_IMX_HAB_CST_KEY_SIZE` | For **RSA** keys, this would be the key length (in bits) as entered into the CST tool. For **ECDSA**, this would be a string determined from the generated certificate file name; for example, for a file named `SRK1_sha256_secp384r1_v3_ca_crt.pem` (found in the certificates directory) the present variable would be set to `secp384r1`. | `2048` |
+| `TDX_IMX_HAB_CST_KEY_EXP` | Key exponent for RSA keys (only). | `65537` |
 | `TDX_IMX_HAB_CST_DIG_ALGO` | Digest algorithm as entered into the CST tool. | `sha256` |
 | `TDX_IMX_HAB_CST_SRK_CA` | Whether or not the SRK certificates have the CA flag set as entered into the CST tool; allowed values: `0` or `1`. | `1` |
+| `TDX_IMX_HAB_CST_SRK_INDEX` | Zero-based index of the SRK to be used for signing within the SRK table; allowed values: `0`..`3` (corresponding to `SRK1`..`SRK4`, respectively). | `0` |
```
jsrc27: Would it be possibly better if `TDX_IMX_HAB_CST_SRK_INDEX` was index 1 instead of index 0?

The reason I suggest this, is because I can see a potential UX issue/confusion for users. The filenames of the SRKs are index 1. There could be potential miscommunications where customers want to use the "2nd" SRK.  A user who is not careful could set this variable to "2" instead of the correct "1".

I don't think this is a major concern all said. Though I just wanted to raise this possible scenario I imagined. I do see changing this to index 1 would require changes in a few other locations throughout the code as well. So it's fine for me if we decide to keep it index 0.

rborn-tx: @jsrc27 I've used a zero-based index because such index would match the value that goes into the CSF file. But I'd agree that this is exposing low-level details to the user which is unnecessary.

Anyways, I don't have a strong opinion on this topic. However, now that you ask, yeah I believe  having the index match the number that appears in the file names might be a bit more natural. On the implementation side, changing it shouldn't be hard either...

If we want to do that, this is the right moment because we're also changing the name of the variable and throwing an error if the old variable is used. So, we can take the opportunity to also let users know that the indexing also changed.

Just let me know.

@sergioprado Any opinion on this?

sergioprado: Yeah, it's really frustrating that NXP uses SRK 1-4 consistently across their documentation, tooling, and certificate file naming, but then chose to use a 0-indexed value in the CST file.

While I don't have a strong opinion on this, I would suggest using 1-4 for the variable, as it aligns better with the documentation and, particularly, with the naming convention of the certificate files.

I would also recommend documenting this explicitly in the README and adding a note in the code where the variable is initialized to clarify the rationale.

rborn-tx: Ok, in the latest version the index is 1-based.

I've added a comment about the indexing with the in-code comments but I preferred not to say anything about the choice of 1-based indexing on the user documentation as I didn't see much benefit to users in exposing them to that information. But if anyone disagrees, please let me know.

classes/imx-hab.bbclass
```diff
@@ -76,19 +77,19 @@ def make_sub_cert_name(d, prefix, basedir):
...
+        res = f"{prefix}{kidx}_1_{dalgo}_{ksize}_{kexp}_v3_usr_crt.pem"
     elif crypto == "ecdsa":
         if ksize.isdigit():
             bb.warn("TDX_IMX_HAB_CST_KEY_SIZE is likely not set correctly;"
                     "check the documentation to understand how to set this variable for ECDSA.")
-        res = f"{prefix}{kidx+1}_1_{dalgo}_{ksize}_v3_usr_crt.pem"
+        res = f"{prefix}{kidx}_1_{dalgo}_{ksize}_v3_usr_crt.pem"
     else:
         bb.fatal('TDX_IMX_HAB_CST_CRYPTO is not set correctly'
                  '(its value must be either "rsa" or "ecdsa")')
```
sergioprado, line 102: Changes here look good. Just found a typo in the commit message: configuation -> configuration

sergioprado (approved):
Just found a small typo in a commit message. Despite that, looks good to me.

classes/imx-hab.bbclass
rborn-tx: Updated. Thanks!

jsrc27 (approved):

PR #118, opened 2025-05-21
Attempt to fix error "Customer key cannot be found" when building signed verdin-am62 images

rborn-tx:
This is an attempt to solve a build failure happening on Jenkins in which the logs show this:

> Error: Customer key cannot be found, correctly define TI_SECURE_DEV_PKG
> environment variable

Originally the code was creating a symlink from the source directory to the keys directory passed by the user which ended up creating a sysroot with symlinks pointing outside the sysroot. Normally a sysroot is expected to be a self-contained environment, thus having such symlinks would not be ideal.

With this commit we copy the key files directory into the source directory which in turn makes the sysroot self contained. The hypothesis is that the external symlinks could be causing issues with the sstate mechanism and removing them would solve the issue in CI. However, this must be confirmed since the issue in CI couldn't be reproduced locally.

Related-to: TOR-3840

### DIFF
--- a/dynamic-layers/meta-ti-bsp/recipes-ti/secdev/ti-k3-secdev-secboot.inc
+++ b/dynamic-layers/meta-ti-bsp/recipes-ti/secdev/ti-k3-secdev-secboot.inc
@@ -1,6 +1,7 @@
 update_signing_keys() {
-    rm -Rf ${S}/keys
-    ln -s ${TDX_K3_SECBOOT_KEY_DIR} ${S}/keys
+    echo "Updating signing keys directory '${S}/keys'"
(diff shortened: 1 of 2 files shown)


sergioprado (approved):
@rborn-tx Changes look good to me, so I am approving this.

Since we are doing the same change in two different places, I was just wondering if we should squash it into a single commit or not. I will leave that up to you.

rborn-tx:
> Since we are doing the same change in two different places, I was just wondering if we should squash it into a single commit or not. I will leave that up to you.

@sergioprado Since the commits affect different recipes, I'm leaning towards keeping them separate.

jsrc27 (approved):

PR #120, opened 2025-05-23
u-boot: show CST output upon errors

rborn-tx:
Previously when CST failed to execute no error messages were available in the task log which made hard for users to figure out the reason of the failure. With this commit we dump the output of the tool execution to improve the situation.


sergioprado (approved):
LGTM

jsrc27 (approved):
LGTM

PR #121, opened 2025-05-29
u-boot-hab.inc: use libfaketime when generating CSF binaries

lucas-akira:
When generating the CSF binaries, the NXP Code Signing Tool adds the current UTC timestamp in string format to them, resulting in files that are not reproducible. It is possible to circumvent this limitation by using a library like libfaketime during the build process to trick CST to always add the same timestamp to the binaries.

Related-to: TOR-3842

### DIFF
--- a/recipes-bsp/u-boot/files/imx8m_sign.sh
+++ b/recipes-bsp/u-boot/files/imx8m_sign.sh
@@ -9,6 +9,12 @@ UBOOT_SPL_DDR_BINARY="${UBOOT_SPL_DDR_BINARY:-u-boot-spl-ddr.bin}"
 UBOOT_DTB_BINARY="${UBOOT_DTB_BINARY:-u-boot.dtb.out}"
 UBOOT_CONTAINER_BINARY="${UBOOT_CONTAINER_BINARY:-flash.bin}"
 CSF_PREFIX="${CSF_PREFIX:-csf-for-}"
+LIBFAKETIME_PATH="${LIBFAKETIME_PATH}"
+
+# Timestamp to be used by libfaketime. The date itself isn't important, but this needs
+# to be constant between builds in order to generate deterministic CSF binaries
+# Default value is: November 6, 2024 (12:00 PM UTC)
+FAKETIME=${FAKETIME:-1730894400}
 
 # shellcheck disable=SC2155
 readonly FILE_SCRIPT="$(basename "$0")"
@@ -52,6 +58,8 @@ help() {
     echo "                              default: flash.bin"
     echo "    CSF_PREFIX                Prefix for CSF files to be produced"
     echo "                              default: csf-for-"
+    echo "    LIBFAKETIME_PATH          Path to libfaketime library (needed for reproducible CSF binary builds)"
+    echo "                              default: unset"
     echo
     echo " Optional switches:"
     echo "    -h --help            display this Help message"
@@ -209,7 +217,11 @@ generate_spl_csf_and_update_container() {
     sed -i "/Blocks = / s@.*@    Blocks = ${spl_block_base} 0x0 ${spl_block_size} \"${UBOOT_CONTAINER_BINARY}\"@" "${CSF_SPL}.csf"
 
     # Generate CSF blob:
-    if ! "${TDX_IMX_HAB_CST_BIN}" -i "${CSF_SPL}.csf" -o "${CSF_SPL}.bin" > "${CSF_SPL}.log" 2>&1; then
+    if ! env ${LIBFAKETIME_PATH+LD_PRELOAD="${LIBFAKETIME_PATH}"
+                                FAKETIME_FMT="%s"
+                                FAKETIME="@${FAKETIME}"} \
+             "${TDX_IMX_HAB_CST_BIN}" -i "${CSF_SPL}.csf" -o "${CSF_SPL}.bin" > "${CSF_SPL}.log" 2>&1
+    then
         echo "CST execution log:" >&2
         cat "${CSF_SPL}.log" | sed 's@^@|@' >&2
         error "CST failed to execute; please check logs."
@@ -278,7 +290,11 @@ generate_fit_csf_and_update_container() {
     dd if=ivt.bin of="${UBOOT_CONTAINER_BINARY}" bs=1 seek=${ivt_block_offset} conv=notrunc
 
     # Generate CSF blob:
-    if ! "${TDX_IMX_HAB_CST_BIN}" -i "${CSF_FIT}.csf" -o "${CSF_FIT}.bin" > "${CSF_FIT}.log" 2>&1; then
+    if ! env ${LIBFAKETIME_PATH+LD_PRELOAD="${LIBFAKETIME_PATH}"
+                                FAKETIME_FMT="%s"
+                                FAKETIME="@${FAKETIME}"} \
+             "${TDX_IMX_HAB_CST_BIN}" -i "${CSF_FIT}.csf" -o "${CSF_FIT}.bin" > "${CSF_FIT}.log" 2>&1
+    then
         echo "CST execution log:" >&2
         cat "${CSF_FIT}.log" | sed 's@^@|@' >&2
         error "CST failed to execute; please check logs."
--- a/recipes-bsp/u-boot/u-boot-hab.inc
+++ b/recipes-bsp/u-boot/u-boot-hab.inc
@@ -6,6 +6,9 @@ DEPENDS += "util-linux-native"
 # Need this for the xxd and fdtget utilities used by imx8m_sign.sh
 DEPENDS += "xxd-native dtc-native"
 
+# Need this to create deterministic CSF binaries when running the NXP Code Signing Tool
+DEPENDS += "libfaketime-native"
+
 SRC_URI:append = "\
     file://u-boot-hab.cfg \
     file://create_fuse_cmds.sh \
@@ -49,6 +52,7 @@ imx6_imx7_sign_habv4() {
 
     bbdebug 1 "Generating CSF for U-Boot"
     # Generate CSF file:
+    LIBFAKETIME_PATH="${STAGING_LIBDIR_NATIVE}/faketime/libfaketime.so.1" \
     TDX_IMX_HAB_CST_SRK="${TDX_IMX_HAB_CST_SRK}" \
     TDX_IMX_HAB_CST_SRK_CERT="${TDX_IMX_HAB_CST_SRK_CERT}" \
     TDX_IMX_HAB_CST_CSF_CERT="${TDX_IMX_HAB_CST_CSF_CERT}" \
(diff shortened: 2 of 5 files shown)


recipes-test/libfaketime/libfaketime_%.bbappend
```diff
@@ -0,0 +1 @@
+BBCLASSEXTEND += "native nativesdk"
```
sergioprado: Any reason to have this installed in the SDK?

recipes-bsp/u-boot/u-boot-hab.inc
```diff
@@ -6,6 +6,14 @@ DEPENDS += "util-linux-native"
 # Need this for the xxd and fdtget utilities used by imx8m_sign.sh
 DEPENDS += "xxd-native dtc-native"
 
+# Need this to create deterministic CSF binaries when running the NXP Code Signing Tool
+DEPENDS += "libfaketime-native"
+
+# Timestamp to be used by libfaketime. The date itself isn't important, but this needs
+# to be constant between builds in order to generate deterministic CSF binaries
+CST_FAKE_UNIX_TIMESTAMP = "1730894400"
+# Torizon OS 7.0.0 public announcement date: November 6, 2024 (12:00 PM UTC time arbitrarily chosen)
```
sergioprado: Maybe this comment could be removed? Having Torizon OS in the comment might be confusing, as this is a generic layer. Maybe you could move this comment on top of the variable:

```
# Timestamp to be used by libfaketime. The date itself isn't important, but this needs
# to be constant between builds in order to generate deterministic CSF binaries
# Default value is: November 6, 2024 (12:00 PM UTC)
CST_FAKE_UNIX_TIMESTAMP = "1730894400"
```

sergioprado (changes requested):
I added some initial comments. Also, do we need this for AHAB? The AHAB signing logic is in   `dynamic-layers/freescale/recipes-bsp/imx-mkimage/imx-boot-hab.inc`.

jsrc27:
Just to understand, this timestamp we're manipulating here. It's just simply timestamp metadata in the binary right?

It serves no other purpose?

recipes-test/libfaketime/libfaketime_%.bbappend
lucas-akira: Hmm, I copied this line from another recipe and didn't think too much about it. But good point, I don't see any reason to have the library in the SDK. I'll remove it.

recipes-bsp/u-boot/u-boot-hab.inc
lucas-akira: >  Having Torizon OS in the comment might be confusing, as this is a generic layer.

That makes sense. I'll use your suggestion then, thanks!

lucas-akira:
> Just to understand, this timestamp we're manipulating here. It's just simply timestamp metadata in the binary right?
> 
> It serves no other purpose?

Well, from what I could see by comparing two CSF binaries built without using libfaketime, the timestamp does appear to affect a 256-byte block next to it in the CSF binary as well:

<details><summary>Example: Difference between two CSF binaries built one hour apart on April 28th</summary>
<p>

image

</p>
</details> 

I'm not sure what the block represents, but probably CST uses the timestamp as an input for the signing process.

Inside a CSF binary file I always find two (identical) timestamps at different offsets, and so two 256-byte blocks end up being affected.

If I use libfaketime, both the timestamp and the 256-byte blocks end up being identical between builds.

recipes-bsp/u-boot/u-boot-hab.inc
```diff
@@ -49,6 +57,9 @@ imx6_imx7_sign_habv4() {
 
     bbdebug 1 "Generating CSF for U-Boot"
     # Generate CSF file:
+    LD_PRELOAD="${STAGING_LIBDIR_NATIVE}/faketime/libfaketime.so.1" \
+    FAKETIME_FMT="%s" \
+    FAKETIME="@${CST_FAKE_UNIX_TIMESTAMP}" \
```
rborn-tx: Here we're faking the time for the whole signing script. Have you considered using a fake time only with CST? Or is there a good reason for encompassing the whole script?

rborn-tx (commented):
Being able to generate binaries with the exact same contents is a nice feature!

Overall, looks good. I've raised a question regarding the approach though, which might lead to changes.

lucas-akira:
> I added some initial comments. Also, do we need this for AHAB? The AHAB signing logic is in `dynamic-layers/freescale/recipes-bsp/imx-mkimage/imx-boot-hab.inc`.

I would say probably yes, though I only tested it for the HAB process. 

I'm currently focusing on HAB as I'm also working on implementing the signing feature on TorizonCore Builder which will, for now, only support the Verdin i.MX8M Plus. I can take a look at AHAB after this is done.

recipes-bsp/u-boot/u-boot-hab.inc
lucas-akira: You mean only using the fake time in the actual CST command call inside `imx8m_sign.sh`/`imx6_imx7_create_csf.sh`? I haven't considered it, but it makes sense to restrict its usage to the relevant command only. I'll adapt the scripts instead of the include file then.

recipes-bsp/u-boot/files/imx8m_sign.sh
```diff
@@ -52,6 +54,10 @@ help() {
     echo "                              default: flash.bin"
     echo "    CSF_PREFIX                Prefix for CSF files to be produced"
     echo "                              default: csf-for-"
+    echo "    CST_FAKE_UNIX_TIMESTAMP   Unix timestamp to be used by CST when creating the CSF binaries"
+    echo "                              default: unset (current time will be used)"
+    echo "    LIBFAKETIME_PATH          Path to the libfaketime library (needed if using CST_FAKE_UNIX_TIMESTAMP)"
+    echo "                              default: unset"
```
sergioprado, line 62: @lucas-akira, is there a specific reason for passing these values as parameters to the script? Since you're using the same values every time, it might be more efficient to hardcode them directly into the script. This would simplify the implementation and centralize everything within the script, eliminating the need to modify the recipe. Does that make sense?

lucas-akira: I passed them as optional parameters to keep the script decoupled from the Yocto/OE environment, as it was before, and to keep its general behavior unchanged.

Since the usage of libfaketime is not strictly necessary for the signing process itself, I tried to keep the script behavior the same as before if either the timestamp or the libfaketime path is not explicitly provided.

Of course we can discuss if it's worth doing this since the same values are used every time.

> Since you're using the same values every time, it might be more efficient to hardcode them directly into the script. This would simplify the implementation and centralize everything within the script, eliminating the need to modify the recipe. Does that make sense?

I see your point, though I wonder if it would be better for maintenance to at least define the timestamp constant in one place (i.e. in the recipe) rather than repeating its value for all signing scripts (`imx8m_sign.sh`/`imx6_imx7_create_csf.sh`).

sergioprado: > I passed them as optional parameters to keep the script decoupled from the Yocto/OE environment, as it was before, and to keep its general behavior unchanged.

The other parameters were decoupled from the script because they are configurable and may be changed by the user. As far as I understood, the parameters you are introducing are not configurablw, so I think they can be encapsulated inside the script. This would make the implementation much more simpler, improving maintenance.

Is there any reason why one would want to change one of its parameters, for example the timestamp? If not, them makes more sense to me to have it encapsulated inside the script.

@jsrc27 @rborn-tx What are your thoughts on this?

rborn-tx: I don't have a strong opinion, but since the whole approach of setting a fake time is just a means to an end, namely, to have reproducible builds, I don't see the need to make the fake date/time configurable via Bitbake variables as nobody should have to care about the particular date/time employed (each script could even have its own value btw). Internally in the script we could have a default that could be overridden if someone wanted to run the script standalone and run with different value which could be done like this inside the signing script:

```
FAKETIME=${FAKETIME:-1730894400}
```

As for the path to the `faketime` library, it seems we need to pass it to the script since the information is available on the Yocto side only. Hard-coding the value inside the script would probably be worse because the coupling wouldn't be much visible in that way.

Btw, we could set the variables `LD_PRELOAD`, `FAKETIME_FMT` and `FAKETIME` exclusively when `LIBFAKETIME_PATH` is set. That could be done like this:

```
env ${LIBFAKETIME_PATH+LD_PRELOAD="${LIBFAKETIME_PATH}"
                       FAKETIME_FMT="%s"
                       FAKETIME="@${FAKETIME}"} \
    "${TDX_IMX_HAB_CST_BIN}" -i "${CSF_FIT}.csf" -o "${CSF_FIT}.bin" > "${CSF_FIT}.log" 2>&1
```

jsrc27: I have a similar opinion, if the parameter is not meant to be changed/configured, then it doesn't seem to make sense to have it as an input that looks configurable.

Unless there is some benefit or use-case to having it configurable then that can change things a bit.

Slight tangent, from a UX/transparency standpoint is there any benefit to having this detail be transparent to users? Even just a small footnote in the README, like "we hard-code the timestamp because XYZ". Or do we just leave a in-code comment somewhere and that would suffice for curious users?

sergioprado: So I guess the conclusion is:

- Only pass `LIBFAKETIME_PATH` to the script, and make it optional. If it is not passed, then `libkaketime` it is not used.
- Encapsulate `FAKETIME` inside the script, but make it possible for a user of the script to change it.
- Evaluate the need for adding a comment in the `README` or in the code itself to make sure users understand this is being done to have reproducible builds.

lucas-akira: Right, makes sense that the user shouldn't really care about setting a particular fake timestamp, as the value itself doesn't affect the end goal to have reproducible builds.

So @sergioprado I agree that the timestamp value shouldn't be a parameter for the scripts. There really isn't a very good reason for a user to change it.

> So I guess the conclusion is:
> 
>     * Only pass `LIBFAKETIME_PATH` to the script, and make it optional. If it is not passed, then `libkaketime` it is not used.
> 
>     * Encapsulate `FAKETIME` inside the script, but make it possible for a user of the script to change it.
> 
>     * Evaluate the need for adding a comment in the `README` or in the code itself to make sure users understand this is being done to have reproducible builds.

Sounds good to me. @rborn-tx I'll follow your suggestions as well.

@jsrc27 I plan to add an in-code comment explaining the library usage before the `FAKETIME` variable declaration, just as I did for the current version that has the value in the recipe. I can add a comment in the iMX secure boot README as well briefly explaining that we use libfaketime to create reproducible bootloader containers.

sergioprado (approved):
Nice! Looks good to me.

jsrc27 (approved):

rborn-tx (approved):
Nice!

PR #123, opened 2025-07-02
Add OP-TEE support for Colibri iMX6

sergioprado:
Build and runtime tested on Colibri iMX6DL 512M.

Commits in this PR:

11849d317eeb optee: enable support for Colibri iMX6
a7d2756f1037 u-boot: generate image with OP-TEE for Colibri iMX6
c807543cdd31 linux: add OP-TEE DT nodes when TDX_OPTEE_DT_OVERLAY is enabled
e962723c9007 tdx-optee: include machine-specific fixups


classes/include/optee/linux-dt-overlay.inc
```diff
@@ -0,0 +1,5 @@
+# generate device tree overlay with OP-TEE nodes
+TDX_OPTEE_DT_OVERLAY = "1"
```
jsrc27, line 2: Just to try and understand this variable more. So for imx6 we always want this to be `1` if we want to use OPTEE correct?

And this is more on a per-machine basis whether this should be set or not correct?

jsrc27 (approved):
LGTM, just one minor question.

classes/include/optee/linux-dt-overlay.inc
sergioprado: @jsrc27 Thanks for reviewing!

> Just to try and understand this variable more. So for imx6 we always want this to be 1 if we want to use OPTEE correct?

The idea is that machines requiring OP-TEE-related device tree nodes should include `linux-dt-overlay.inc`, which sets `TDX_OPTEE_DT_OVERLAY`. This triggers the generation of a device tree overlay that will be loaded at boot time. Currently, all i.MX6 and i.MX7-based SoMs will need this.

> And this is more on a per-machine basis whether this should be set or not correct?

Exactly. It depends on the device tree used by the specific machine. If the default device tree doesn't include the necessary OP-TEE nodes, then enabling this is required.

recipes-bsp/u-boot/u-boot-optee.inc
```diff
@@ -9,4 +9,54 @@ SRC_URI:append:verdin-imx8mm = "\
...
+do_prepare_uboot_for_optee[depends] += "optee-os:do_deploy"
+addtask do_prepare_uboot_for_optee before do_compile after do_configure
+
+do_configure:append:mx6-generic-bsp() {
+    # add binman node to U-Boot device tree
+    sed -i 's|@@TDX_OPTEE_ADDR_LOAD@@|${TDX_OPTEE_ADDR_LOAD}|g' ${WORKDIR}/imx-v6-v7-optee.dtsi
+    cp ${WORKDIR}/imx-v6-v7-optee.dtsi ${S}/arch/arm/dts/
+    for soc in imx6qdl imx6ull; do
+        echo '#include "imx-v6-v7-optee.dtsi"' >> ${S}/arch/arm/dts/$soc-colibri.dtsi
+    done
```
rborn-tx: @sergioprado Have you considered the use of `CONFIG_DEVICE_TREE_INCLUDES` to achieve the goal of this appending?

recipes-bsp/u-boot/u-boot-optee.inc
```diff
@@ -9,4 +9,54 @@ SRC_URI:append:verdin-imx8mm = "\
...
+
+    # make sure binman is selected for iMX6
+    sed -i '/config MX6$/a\\tselect BINMAN' ${S}/arch/arm/mach-imx/mx6/Kconfig
```
rborn-tx, line 56: @sergioprado Have you considered setting `CONFIG_BINMAN` through a `.cfg` file instead?

Btw, it's awesome that we can create the U-Boot binary through BINMAN.

rborn-tx (commented):
That's a nice addition.

I've raised some questions for your evaluation.

recipes-bsp/u-boot/u-boot-optee.inc
sergioprado: Hmmm, interesting. I was not aware of this config and it might simplify the implementation. I will look into it.

recipes-bsp/u-boot/u-boot-optee.inc
sergioprado: Hmmm, yeah. That was the first thing I tried, but it didn't work.

This is because `CONFIG_BINMAN` does not have a prompt:

```
config BINMAN
        bool
        select DTOC
```

And according to the [Kconfig especification](https://docs.kernel.org/kbuild/kconfig-language.html):

```
If a prompt is not present, the config option is a non-visible symbol, meaning its value
cannot be directly changed by the user (such as altering the value in .config) and the
option will not appear in any config menus. Its value can only be set via “default” and
“select” (see below).
```

Since I didn't want to create a patch for a one line change, sed'ing was the best solution I could find out. But I am open to other suggestions.

rborn-tx: Got it. Yeah that looks like the best we can do.

recipes-bsp/u-boot/u-boot-optee.inc
sergioprado: @rborn-tx Using `CONFIG_DEVICE_TREE_INCLUDES` worked like a charm. Thanks for the suggestion. The MR was updated and it is ready for another review.

rborn-tx: Awesome!

rborn-tx (approved):

rborn-tx:
@sergioprado We're in feature freeze due to the Torizon OS quarterly release, so let's wait a bit before merging this, ok?

sergioprado:
@rborn-tx Sure. What is the expected release date?

rborn-tx:
@sergioprado We don't have a precise date but it seems it'll take longer than usual because the release will include a couple of new machines for which even the tests are not fully settled.

sergioprado:
@jsrc27 @rborn-tx FYI, I will merge this into `scarthgap-next`.

sergioprado:
Alright. So after talking to Jon, I will hold the merge, as he thinks we should merge it directly on `scarthgap-7.x.y`. I will start some internal discussion on this.

sergioprado:
After some internal discussion, we will move on with the `next` branch strategy. I am merging this now.

PR #124, opened 2025-07-08
Adapt the hardening for use with "Standard Boot / bootflow"

rborn-tx:
These are additions to allow boot scripts to use the "bootflow" command for booting devices when the hardening is enabled (in particular the command whitelist feature which is part of the hardening).

Basically this adds the various "bootflow" sub-commands to the whitelist with appropriate categories.

Besides adding the (sub-)commands to the whitelist, here we add logic to detect at build-time whether unsupported bootmeths are being enabled in U-Boot. If so, the build will fail forcing users to review their configuration.

Additionally, a variable can be set to cause a config fragment to be applied to U-Boot disabling most bootmeths. At the moment, this variable is set such that the fragment is not applied. The idea is that we could set the variable differently per machine depending on whether the machine has already switched to bootflow (Standard Boot).

For details, please refer to the individual commit messages.

Related-to: TOR-3865


docs/README-secure-boot.md
```diff
@@ -73,6 +73,8 @@ The hardening features are controlled by the variables in the following table pl
 | `TDX_UBOOT_HARDENING_ENABLE_DBG` | Enable debugging of the hardening features; among other things this adds a command to U-Boot that allows the hardening and the device state (open/closed in terms of secure boot) to be manually set for testing purposes - **never** do this in production | `0` |
 | `TDX_SECBOOT_REQUIRED_BOOTARGS` | Expected value for the fixed part of the kernel command line | Different value for each machine |
 | `TDX_AMEND_BOOT_SCRIPT` | When set to `1` the boot script will be amended to make it suitable for secure boot; this only works with the script provided by Toradex for BSP reference images; users employing a custom script should set this to `0` | Same value as variable `TDX_UBOOT_HARDENING_ENABLE` |
+| `TDX_CHECK_UNSUPP_BOOTMETH` | Enable a build-time check to determine if an unsupported [bootmeth](https://docs.u-boot.org/en/stable/develop/bootstd/overview.html#bootmeth) is being enabled in U-Boot (which could become an attack vector). | `1` |
```
jsrc27, line 76: Is there a reason a user would not want this? Or is this variable just purely to give the user a choice if they want?

jsrc27 (commented):
Left a minor question/comment.

recipes-bsp/u-boot/u-boot-harden.inc
```diff
@@ -99,6 +99,50 @@ amend_secboot_dtsi() {
...
+"
+check_dotconfig_compat_all() {
+    if [ -n "${UBOOT_CONFIG}" ]; then
+        unset i j
+        for config in ${UBOOT_MACHINE}; do
+            i=$(expr $i + 1);
+            for type in ${UBOOT_CONFIG}; do
+                j=$(expr $j + 1);
+                if [ $j -eq $i ]; then
+                    check_dotconfig_compat $config $type
```
sergioprado, line 114: It seems `$type` is not being used in `check_dotconfig_compat`.

recipes-bsp/u-boot/u-boot-harden.inc
```diff
@@ -99,6 +99,50 @@ amend_secboot_dtsi() {
...
+        done
+        unset i
+    else
+        check_dotconfig_compat
+    fi
+}
+
+check_dotconfig_compat() {
+    config=$1
+    #type=$2
```
sergioprado: Can this commented line be removed?

sergioprado (commented):
Looks good to me overall. Just left a couple of comments.

recipes-bsp/u-boot/u-boot-harden.inc
rborn-tx: That's right. But I kept it like that because the outer function (`check_dotconfig_compat_all()`) is basically boiler plate code (dealing with the use/non-use of `UBOOT_CONFIG`) while the actual work is done by the inner function (`check_dotconfig_compat()`). At the moment the inner function happens to not use `type` but other layers could replace or append that function and perhaps use that parameter, considering a generic scenario.

recipes-bsp/u-boot/u-boot-harden.inc
rborn-tx: Sure. Done on v2.

docs/README-secure-boot.md
rborn-tx: Since this check stops the build with a fatal error, I thought it was better to have a way to disable it just in case users want to bypass that check. Of course, if they do that they would be assuming the risk of allowing a bootmeth that is not protected by the hardening.

Notice that the only bootmeth supported by the hardening is the one called "script" which is also the one used for booting Torizon OS and the BSP reference images. All the other bootmeths are unsupported but not necessarily insecure. Therefore, there might be valid and even secure use cases where someone could enable other bootmeths.

sergioprado (approved):

jsrc27 (approved):

PR #125, opened 2025-07-09
u-boot-harden: disable unsupported bootmeths by default

rborn-tx:
Set default value of TDX_DISABLE_UNSUPP_BOOTMETH from "0" to "1" to avoid breaking the build on machines that are already enabling "Standard Boot" intentionally or not.

Related-to: TOR-3865


sergioprado (approved):
Yeah, I confirm without this the build for Colibri iMX6 fails because it has "Standard Boot" enabled by default.

LGTM.

rborn-tx:
Guys, I'll go ahead and merge this to see if we can avoid breaking the nightly signed builds.

PR #126, opened 2025-07-16
optee: enable support for Apalis iMX6

sergioprado:
This merge request enables OP-TEE support for Apalis iMX6.

Because most implementation is shared with Colibri iMX6, a few overrides are renamed from `colibri-imx6` to `mx6-generic-bsp`.

Build and runtime tested on Apalis iMX6Q 1GB V1.1B.


rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #127, opened 2025-07-17
Enable the hardening features on the aquila-am69

rborn-tx:
With "Standard Boot" being now supported by the hardening, in particular the "bootflow" command used by newer Toradex modules such as the aquila-am69, enable the hardening on that SoM.

Related-to: TOR-3886

### DIFF
--- a/classes/include/signed/tdx-signed-harden.inc
+++ b/classes/include/signed/tdx-signed-harden.inc
@@ -13,7 +13,7 @@
 # UBOOT_SIGN_ENABLE are set, by default.
 #
 def default_tdx_uboot_hardening_enable(d):
-    unsupported_machines = ["aquila-am69", "imx95-19x19-verdin"]
+    unsupported_machines = ["imx95-19x19-verdin"]
     for machine in unsupported_machines:
         if (machine in d.getVar('OVERRIDES').split(':')):
             bb.warnonce("%s machine requires some additional work to have "\


sergioprado:
@rborn-tx JUst to confirm, this was build and runtime tested, right?

jsrc27 (approved):

rborn-tx:
> @rborn-tx JUst to confirm, this was build and runtime tested, right?

@sergioprado Actually, this was only build tested. Since I don't have that hardware at hand I decided to take that risk after checking the BCoT tests on LAVA for that machine where most of the tests are failing due to the lack of the hardening.

PS.: We've just merged some changes to the BCoT tests in LAVA to allow them to work when bootflow is in use, such as in case of the `aquila-am69`.

sergioprado:
> @sergioprado Actually, this was only build tested. Since I don't have that hardware at hand I decided to take that risk after checking the BCoT tests on LAVA for that machine where most of the tests are failing due to the lack of the hardening.

Understood. I'd happily test it myself, but unfortunately, I'll be unavailable for about a week and don't have access to the hardware right now.

Let's proceed with merging, and we can address any issues if they arise in CI.

Also, you're probably aware, but since we're merging this into `scarthgap-next`, CI will only pick up these changes once they're merged into `scarthgap-7.x.y`.

PR #130, opened 2025-08-06
u-boot: use downstream patches on the toradex-smarc-imx95 machine

rborn-tx:
Select the downstream secure-boot patches on the toradex-smarc-imx95 since that machine uses downstream U-Boot (while the default patches are for upstream).

Related-to: TOR-3888


sergioprado (approved):
LGTM.

rborn-tx:
@sergioprado Ok, I'll go ahead and merge this since the other reviewer is unavailable.

PR #132, opened 2025-08-19
Properly detect open/closed state of the i.MX95 devices

rborn-tx:
- Add support for detecting the current life-cycle state of an i.MX95 device (impacting all hardening features).
- Set expected bootargs for the toradex-smarc-imx95.

Related-to: TOR-3932

### DIFF
--- a/recipes-bsp/u-boot/u-boot-secure-boot.inc
+++ b/recipes-bsp/u-boot/u-boot-secure-boot.inc
@@ -13,6 +13,7 @@ TDX_UBOOT_SECBOOT_PATCHES_UPSTREAM = " \
 # NOTE: NXP downstream does not need patch "lmb: prevent alloc overlap with reserved regions"
 TDX_UBOOT_SECBOOT_PATCHES_DOWNSTREAM = " \
     ${TDX_UBOOT_SECBOOT_PATCHES_COMMON} \
+    file://0001-toradex-add-helper-to-detect-closed-iMX95-downstream.patch \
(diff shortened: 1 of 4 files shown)


recipes-bsp/u-boot/u-boot-secure-boot.inc
```diff
@@ -13,6 +13,7 @@ TDX_UBOOT_SECBOOT_PATCHES_UPSTREAM = " \
 # NOTE: NXP downstream does not need patch "lmb: prevent alloc overlap with reserved regions"
 TDX_UBOOT_SECBOOT_PATCHES_DOWNSTREAM = " \
     ${TDX_UBOOT_SECBOOT_PATCHES_COMMON} \
+    file://0001-toradex-add-helper-to-detect-closed-iMX95-downstream.patch \
```
sergioprado, line 16: I see that this patch will apply to all downstream U-Boot machines, including Apalis iMX8 and Colibri iMX8X. Are the U-Boot sources for those machines the same as the ones used on iMX95? I just want to make sure the patch applies cleanly across all downstream U-Boot versions.

rborn-tx: Good point. Yes, all downstream NXP recipes are using U-Boot v2024.04 such that the new patch applies cleanly.

sergioprado: Perfect, thanks for confirming!

sergioprado (approved):
LGTM.

PR #133, opened 2025-08-27
Fix dependency issue when dm-verity is enabled 

sergioprado:
Fix the following build error:

```
08-21:18:57:05  ERROR: tdx-reference-minimal-image-1.0-r0 do_image_bootfs: Command 'install -m 0644 -D /workdir/oe/tmp-glibc/deploy/images/apalis-imx8/fitImage /workdir/oe/tmp/work/apalis_imx8-tdx-linux/tdx-reference-minimal-image/1.0/bootfs/fitImage' returned 1:
08-21:18:57:05  b"install: cannot stat '/workdir/oe/tmp-glibc/deploy/images/apalis-imx8/fitImage': No such file or directory\n"
```

This was only reproducible in CI, and only on Apalis iMX8 and Colibri iMX8X. It seems there is some weird dependency issue when building a Tezi image with `dm-verity` enabled for those machines. The `do_image_bootfs` task from the Tezi class runs before `do_deploy` from the kernel recipe, so the `fitImage` is not yet available.

Fix it by setting the correct dependency:

```
tdx-reference-minimal-image:do_image_bootfs -> virtual/kernel:do_deploy
```


jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

PR #134, opened 2025-09-03
tdx-signed-harden: set expected bootargs for colibri-imx7-emmc

rborn-tx:
Set expected bootargs for the colibri-imx7-emmc machine (before, they were wrongly set for the colibri-imx7 instead); this impacts mainly BSP images.


sergioprado (approved):
LGTM

PR #135, opened 2025-09-08
optee-client: fix build failure with NXP downstream recipe

sergioprado:
Fixes the following error during do_install:

09-05:18:57:01  | install: cannot stat '/tee-supplicant@.service': No such file or directory
09-05:18:54:26  NOTE: recipe optee-client-4.4.0.imx-r0: task do_install: Failed

The issue occurs because the downstream optee-client recipe from meta-freescale uses the UNPACKDIR variable, which was only introduced in Styhead.

Fix it by initializing UNPACKDIR with WORKDIR.

### DIFF
--- a/recipes-security/optee/optee-client_%.bbappend
+++ b/recipes-security/optee/optee-client_%.bbappend
@@ -4,6 +4,12 @@ SRC_URI += "\
     file://tee-supplicant.rules \
 "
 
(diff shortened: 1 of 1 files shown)


rborn-tx (approved):
LGTM

jsrc27 (approved):
LGTM

PR #144, opened 2025-10-13
Allow passing extra args to NXP CST tool

### LINKED ISSUE #139: HSM use with NXP HAB CST
I would like to use an HSM (specifically YubiHSM2 or SoftHSM to test) with the CST tool initially for the imx8mp and eventually with the imx95.  

In section 7.3 of the UG10106 document it appears this is possible by using the -b pkcs11 parameter of the cst command (page 70).  However searching through the files for the use of TDX_IMX_HAB_CST_BIN I don't see any option to use the -b parameter.  

I would like to request/greatly appreciate this feature.

sergioprado:
Some use cases require passing additional options to the NXP CST tool. For example, when signing with an HSM, the --backend (pkcs11) parameter must be supplied.

Add a new variable, TDX_IMX_HAB_CST_ARGS, which is empty by default and is appended to the cst invocation. This keeps current behaviour unchanged while allowing users to supply extra arguments when needed.

Closes #139


jsrc27 (approved):
LGTM

dynamic-layers/freescale/recipes-bsp/imx-mkimage/files/mx8_sign.sh
```diff
@@ -161,7 +165,7 @@ generate_csf_ahab() {
     echo "CSF location: '${image_csf}'"
 
     # Sign:
-    "${TDX_IMX_HAB_CST_BIN}" -i "${image_csf}" -o "${UNSIGNED_IMAGE}-signed"
+    "${TDX_IMX_HAB_CST_BIN}" "${TDX_IMX_HAB_CST_ARGS}" -i "${image_csf}" -o "${UNSIGNED_IMAGE}-signed"
```
rborn-tx: @sergioprado I believe the quotes surrounding `${TDX_IMX_HAB_CST_ARGS}` are not wanted here, right? With the quotes all arguments would be passed as a single argument to CST which is probably not intended.

Same goes for the other scripts (`imx6_imx7_create_csf.sh`, `imx8m_sign.sh`).

Btw, the removal of the quotes would add a limitation in that no spaces would be allowed in individual arguments being passed to CST. E.g. this wouldn't work:

```
TDX_IMX_HAB_CST_ARGS="--arg 123 --keyname "my key name""
```

But looking at the arguments accepted by CST, it doesn't seem this limitation would cause any trouble for users so it's probably not worth trying to deal with it.

rborn-tx (commented):
Raised a concern/issue that we may want to solve...

Apart from that, everything else looks good to me.

dynamic-layers/freescale/recipes-bsp/imx-mkimage/files/mx8_sign.sh
sergioprado: Good catch! PR updated.

rborn-tx (approved):

PR #145, opened 2025-10-16
Fixes related to the encryption feature

### LINKED ISSUE #97: Encryption Fails After Restarting Daemon on Aquila AM69
I enabled encryption on Aquila AM69 by setting the following variables in `local.conf`:

```
INHERIT += "tdx-tezi-data-partition tdx-encrypted"
TDX_ENC_KEY_BACKEND:forcevariable = "cleartext"
TDX_ENC_KEY_LOCATION = "filesystem"
TDX_ENC_STORAGE_LOCATION = "/dev/mmcblk0p3"
TDX_ENC_STORAGE_MOUNTPOINT = "/data"
TDX_TEZI_DATA_PARTITION_AUTOMOUNT = "0"
```

On the first boot, encryption works as expected. However, if the encryption daemon (`tdx-enc-handler`) is stopped and restarted multiple times, encryption stops working:

```
$ systemctl stop tdx-enc-handler && systemctl start tdx-enc-handler
```

The kernel logs show multiple EXT4 errors, and mounting the partition fails:

```
[   76.109837] EXT4-fs (dm-0): ext4_check_descriptors: Block bitmap for group 216 not in group (block 8689095025519805591)!
[   76.123241] EXT4-fs (dm-0): group descriptors corrupted!
mount: /data: mount(2) system call failed: Structure needs cleaning.
       dmesg(1) may have more information after failed mount system call.
cleartext: ERROR: Could not mount encrypted partition!
```

Some additional information:

- Issue is AM69-specific - It is not reproducible on other modules.
- Storage type does not affect the issue - Happens with both eMMC and USB storage.
- Filesystems do not affect the issue - Tested with EXT4, F2FS, and BTRFS, all exhibiting similar behavior.
- Key storage backend does not affect the issue - Occurs with both `cleartext` and `tpm` backends.
- Kernel version does not affect the issue – The issue persists even after backporting from kernel 6.6 to 6.1, ruling out a regression.

This appears to be a kernel-related issue that requires further debugging, possibly related to dm-crypt interaction with the AM69 platform, incorrect reinitialization of the encryption layer upon daemon restart or potential hardware or driver-specific bug affecting the encryption stack.

sergioprado:
This PR has some fixes related to the encryption feature.

* Avoid `systemd-mount` from automounting the data partition, which was preventing the `tdx-enc` script from preparing the encrypted partition at the first boot.
* Disable HW crypto offload on all K3 platforms, which is causing issues when using dm-crypt.

Build and runtime tested on Aquila AM69, and Verdin AM62(P)

Closes #97

### DIFF
--- a/recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
+++ b/recipes-core/tdx-enc-handler/tdx-enc-handler_1.0.bb
@@ -76,9 +76,9 @@ do_install() {
     fi
 }
 
-# HW offload is not working properly on Verdin AM62P when trying to mount a partition
+# HW offload is not working properly on K3 platforms when trying to mount a partition
 # with dm-crypt, so let's disable it as a workaround for now.
-do_install:append:verdin-am62p() {
+do_install:append:k3() {
     install -d ${D}${sysconfdir}/modprobe.d/
     echo "blacklist sa2ul" > ${D}${sysconfdir}/modprobe.d/sa2ul.conf
 }
--- a/recipes-core/udev/udev-extraconf_%.bbappend
+++ b/recipes-core/udev/udev-extraconf_%.bbappend
@@ -1,0 +1,3 @@
+do_install:append:tdx-encrypted() {
+    echo ${TDX_ENC_STORAGE_LOCATION} >> ${D}${sysconfdir}/udev/mount.ignorelist
+}


jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

PR #147, opened 2025-10-27
Set signing class name

rborn-tx:
Set variable `TDX_SIGNING_CLASS` to the name of the signing class. The idea is to make that information available to other layers so they can add it to the OS image name/description (as that present in the Toradex Easy Installer `image.json` file); this would allow users to tell signed/non-signed images apart.


jsrc27 (approved):
LGTM

sergioprado (approved):
LGTM

PR #148, opened 2025-11-05
Refactor the generation of the fuse-cmds.txt file

sergioprado:
This is refactoring the generation of the `fuse-cmds.txt` file.

Now the `create_fuse_cmds.sh` script is deployed in a separate package so it can be used by different recipes (`u-boot`, `imx-boot`, etc).

A group of template files are packaged together with the script, and they are used to create the `fuse-cmds.txt` file. This simplified a lot the implementation, making it very generic.

Alongside `fuse-cmds.txt`, the build also generates `fuse-info.txt`. This file contains the same fusing configuration, but in a machine-readable format, intended for programmatic consumption by other tools or recipes.

Also, introduce a new variable called `TDX_IMX_HAB_GEN_UBOOT_FUSING_CMD`. It is disabled by default, but if enabled, it will add a function called `prog_secure_boot_fuses` to U-Boot's environment to program the secure boot fuses.

Tested on `apalis-imx8`, `colibri-imx6`, `colibri-imx7-emmc`, `colibri-imx8x`, `verdin-imx8mp` and `verdin-imx95`.

To validate the implementation on all machines, I compared the `fuse-cmds.txt` file generated by the old script with the new one, to make sure they were identical.


jsrc27 (approved):
LGTM no major comments.

Just to make sure I understand though. For the `imx-fuses-native` recipe, from what it looks like, to use the scripts provided by this recipe another recipe just needs `imx-uses-native` in their `DEPENDS` right?

Also it looks like this recipe doesn't need or depend on inheriting any of the signed classes (i.e `tdx-signed`) right? It can just be used independent of those classes.

sergioprado:
> Just to make sure I understand though. For the `imx-fuses-native` recipe, from what it looks like, to use the scripts provided by this recipe another recipe just needs `imx-uses-native` in their `DEPENDS` right?

Yes, then the script and the template files will endup in the native sysroot of the recipe that wants to use them.

> Also it looks like this recipe doesn't need or depend on inheriting any of the signed classes (i.e `tdx-signed`) right? It can just be used independent of those classes.

Exactly.

rborn-tx (commented):
@sergioprado This is a very nice design and implementation improvement to the layer!

In terms of implementation, it all looks good to me. I have just one idea related to naming for your consideration:

Since we're creating a file with a custom machine-readable format, how about defining a custom extension to it?

For example, all files following the new format (templates and the output) could have the `.fuse` extension (or similar) extension.

Also, maybe the templates could have the word `template` in their names.

Something like this:

- `imx6-fuses.cfg` -> `imx6-template.fuse`
- `imx8m-fuses.cfg` -> `imx8m-template.fuse`
- `fuse-info.txt` -> `fuse-info.fuse`

What do you think?

sergioprado:
> Since we're creating a file with a custom machine-readable format, how about defining a custom extension to it?
> 
> For example, all files following the new format (templates and the output) could have the `.fuse` extension (or similar) extension.
> 
> Also, maybe the templates could have the word `template` in their names.
> 
> Something like this:
> 
> * `imx6-fuses.cfg` -> `imx6-template.fuse`
> * `imx8m-fuses.cfg` -> `imx8m-template.fuse`
> * `fuse-info.txt` -> `fuse-info.fuse`
> 
> What do you think?

Yeah, makes total sense. I liked the name for the templates:

`imx6-fuses.cfg` -> `imx6-template.fuse`

For the output, I don't like having `fuse` two times in the name of the file. :-) What do you think about this:

`fuse-info.txt` -> `imx-config.fuse`

rborn-tx:
> For the output, I don't like having fuse two times in the name of the file. :-) What do you think about this:
> 
> fuse-info.txt -> imx-config.fuse

Sounds good to me!

sergioprado:
@rborn-tx I changed the name of the files as we discussed. This is ready for another review.

rborn-tx (approved):
@sergioprado Thanks for making the changes.

rborn-tx:
Btw, I've already created a ticket for us making the changes on `meta-toradex-torizon` to utilize the same source of truth for the fuse addresses.

PR #154, opened 2025-12-30
whinlatter: Minor fixes and adjustments to achieve compatibility

JSydll:
Move a falsely placed append on a TI-specific recipe to the appropriate place as it will break NXP-based configurations otherwise.

Account for the consistent usage of `UNPACKDIR`. Please note that this was only tested on a whinlatter-based build environment.


recipes-security/optee/optee-client_%.bbappend
```diff
@@ -4,18 +4,12 @@ SRC_URI += "\
     file://tee-supplicant.rules \
 "
 
-# UNPACKDIR is a new variable introduced in Styhead, but optee-client recipe
-# from meta-freescale is already using it, and this is breaking the build, so
-# so lets make sure it points to WORKDIR. This can probably be removed as soon
-# as we move to Styhead or later version.
-UNPACKDIR = "${WORKDIR}"
-
```
rborn-tx, line 12: Given the original comment, wouldn't this removal break the build on `scarthgap`?

rborn-tx (commented):
@JSydll Thanks for your contribution.

Since our main/production branch is `scarthgap`, we need to make sure the build works fine on that branch. I'm raising a question related to that...

recipes-security/optee/optee-client_%.bbappend
JSydll: I wasn't able to check this - there was just some tainted memory in the back of my head that the variable was introduced earlier than it was enforced... But we could also opt to change the target branch to something more suitable? The change will definitely be required as soon as the layer moves beyond scarthgap.

JSydll: I wasn't able to check this - there was just some tainted memory in the back of my head that the variable was introduced earlier than it was enforced... But we could also opt to change the target branch to something more suitable? The change will definitely be required as soon as the layer moves beyond scarthgap.

sergioprado: `UNPACKDIR` was introduced in Styhead and not backported to Scarthgap, so unfortunatelly we cannot remove this for now.

conf/layer.conf
```diff
@@ -17,7 +17,7 @@ LAYERRECOMMENDS_meta-toradex-security = "\
     toradex-ti-layer \
 "
 
-LAYERSERIES_COMPAT_meta-toradex-security = "scarthgap walnascar"
+LAYERSERIES_COMPAT_meta-toradex-security = "scarthgap walnascar whinlatter"
```
sergioprado, line 20: This layer currently supports only the latest Yocto Project LTS release, which is Scarthgap.

Walnascar was added solely to run some CI experiments, but since we do not actually support it, it should be actually removed.

The same applies to Whinlatter. As we do not support that release either, we unfortunately cannot add Whinlatter support here.

Our plan is to add support for Wrynose (expected to become the next LTS) as soon as it is officially released.

JSydll:
I'll close the pull request, since the support strategy is now clarified. It may serve as a reference as soon as Wrynose is about to get real.

PR #155, opened 2026-01-07
Adapt secure boot support on iMX95 for B0 silicon

sergioprado:
i.MX95 B0 silicon is based on AHAB v2 and uses the new hybrid container format, so some adaptations are required to support it.

With this change, support for A0/A1 silicon is dropped. This is acceptable, as A0/A1 support has already been removed at the BSP level.

Tested on SMARC iMX95.


sergioprado:
@rborn-tx @jsrc27 This is ready for review.

jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

We could think of renaming `mx8_sign.sh` to `mx8_mx9_sign.sh` to better represent the cases it handles. But, I don't think we need to do it now. So, I'll go ahead and already merge this.

PR #156, opened 2026-01-14
Fix Colibri iMX8X automatic fusing issue

sergioprado:
On some SoCs (e.g. i.MX8X), `fuse prog` prompts for an additional confirmation when targeting ECC-protected fuse words, even when -y is used:

```
Colibri iMX8X # fuse prog -y 0 730 0xBB31FAAE
Programming bank 0 word 0x000002da to 0x18c3f653... Warning: Words in this index range have ECC protection and can only be programmed once per word. Individual bit operations will be rejected after the first one.
 Really program this word? <y/N>
```

This breaks automated fusing in CI and can cause LAVA jobs to time out waiting for user input.

Fix this by setting `force_prog_ecc=y` in U-Boot before running the fusing commands. The variable is only applicable to platforms that implement this behavior (e.g. iMX8X) and it is ignored elsewhere, so setting it unconditionally is safe.


rborn-tx (approved):
LGTM

jsrc27 (approved):

PR #157, opened 2026-01-29
Support signing secure boot images with an HSM

sergioprado:
This extends the existing `tdx-signed` workflow to support signing secure boot images with keys backed by an HSM accessed via PKCS#11.

Tested on Verdin iMX8MM with SoftHSM using the configuration below:

```
TDX_SIGNED_HSM = "1"
TDX_SIGNED_HSM_FIT_TOKEN_URL = "model=SoftHSM%20v2;manufacturer=SoftHSM%20project;serial=f985d09bc4105496;token=fit-hsm"
TDX_SIGNED_HSM_FIT_TOKEN_LABEL = "fit-sign-key"
TDX_SIGNED_HSM_PKCS11_MODULE_PROVIDER = "softhsm-native"
TDX_SIGNED_HSM_PKCS11_MODULE_PATH = "/usr/lib/softhsm/libsofthsm2.so"
TDX_IMX_HAB_CST_SRK_CERT = "pkcs11:token=cst-hsm;object=SRK1_sha256_2048_65537_v3_ca;type=cert;pin-value=${TDX_SIGNED_HSM_TOKEN_PIN}"
TDX_IMX_HAB_CST_CSF_CERT = "pkcs11:token=cst-hsm;object=CSF1_1_sha256_2048_65537_v3_usr;type=cert;pin-value=${TDX_SIGNED_HSM_TOKEN_PIN}"
TDX_IMX_HAB_CST_IMG_CERT = "pkcs11:token=cst-hsm;object=IMG1_1_sha256_2048_65537_v3_usr;type=cert;pin-value=${TDX_SIGNED_HSM_TOKEN_PIN}"
```

Also tested on Verdin iMX8MM with a YubiKey PIV token:

```
TDX_SIGNED_HSM = "1"
TDX_SIGNED_HSM_FIT_TOKEN_URL = "model=YubiKey%20YK5;manufacturer=Yubico%20%28www.yubico.com%29;serial=22863375;token=YubiKey%20PIV%20%2322863375;id=%12"
TDX_SIGNED_HSM_FIT_TOKEN_LABEL = "yk-dev-key"
TDX_SIGNED_HSM_PKCS11_MODULE_PROVIDER = "yubico-piv-tool-native"
TDX_SIGNED_HSM_PKCS11_MODULE_PATH = "/usr/lib/libykcs11.so"
TDX_IMX_HAB_CST_SRK_CERT = "pkcs11:token=YubiKey%20PIV%20%2322863375;id=%06;type=cert;pin-value=${TDX_SIGNED_HSM_TOKEN_PIN}"
TDX_IMX_HAB_CST_CSF_CERT = "pkcs11:token=YubiKey%20PIV%20%2322863375;id=%07;type=cert;pin-value=${TDX_SIGNED_HSM_TOKEN_PIN}"
TDX_IMX_HAB_CST_IMG_CERT = "pkcs11:token=YubiKey%20PIV%20%2322863375;id=%08;type=cert;pin-value=${TDX_SIGNED_HSM_TOKEN_PIN}"
```

Current scope:

- FIT image signing
- HAB-based boot container signing

Planned follow-ups:

- AHAB support
- TI-based SoCs support
- Documentation


sergioprado:
@jsrc27 @rborn-tx This is ready for review.

cc @tkfu

recipes-bsp/u-boot/files/imx8m_sign.sh
```diff
@@ -51,6 +54,14 @@ help() {
     echo " Optional environment variables:"
     echo
     echo "    TDX_IMX_HAB_CST_ARGS      Additional parameters to be passed to the CST tool"
+    echo "    TDX_IMX_HAB_CST_HSM       Enable usage of HSM (via PKCS11) when set to 1"
+    echo "                              default: 0"
+    echo "    TDX_IMX_HAB_CST_SRK_CA    Whether or not the SRK certificates have the CA flag"
+    echo "                              Only used when HSM is enabled (TDX_IMX_HAB_CST_HSM=1)"
+    echo "                              default: 1"
+    echo "    TDX_IMX_HAB_CST_SRK_INDEX Index of the SRK to be used for signing (1..4)"
```
jsrc27, line 62: Question about the usage of this variable. Maybe it will make more sense in the follow-up PRs when the documentation is updated.

In the HSM use-case the signing script consumes this directly to determine which SRK index to use. In the "normal" use-case we still use this variable to determine the index, but in a different way. I assume this is due the abstraction of the certs the use of an HSM presents so the "normal"/previous method doesn't work. Is that correct?

jsrc27 (commented):
Looks good overall, just left a minor comment/question.

recipes-bsp/u-boot/files/imx8m_sign.sh
sergioprado: Exactly. For example. this is how the HSM token URI looks like with the Yubikey:

```
TDX_IMX_HAB_CST_SRK_CERT = "pkcs11:token=YubiKey%20PIV%20%2322863375;id=%06;type=cert;pin-value=${TDX_SIGNED_HSM_TOKEN_PIN}"
```

Because we cannot extract both the CA flag and the key index from the URI, we need to use the variables `TDX_IMX_HAB_CST_SRK_CA` and `TDX_IMX_HAB_CST_SRK_INDEX` for that. I plan to state this in the documentation.

jsrc27 (approved):

rborn-tx (approved):
Great addition!

Btw, this was much cleaner than I was expecting...

PR #159, opened 2026-02-25
[Main] Fix minor build issues on Torizon master branch

EdTheBearded:
Adding layer compatibility with latest Yocto release, as well as renaming a recipe .bbappend that was refactored by BSP.
Details in the commit.


sergioprado:
@EdTheBearded I think the commit title `linux: fix append location for device-tree-overlays-imx` and its message is misleading. In the commit, you say you are `moving the append file to the matching path under ...`, but actually you are just renaming the file. Would you please review?

EdTheBearded:
> @EdTheBearded I think the commit title `linux: fix append location for device-tree-overlays-imx` and its message is misleading. In the commit, you say you are `moving the append file to the matching path under ...`, but actually you are just renaming the file. Would you please review?

Oops, sorry!
Already changed, thanks!

sergioprado (approved):
LGTM.

@rborn-tx Because this is just a minor change with a couple of maintenance commits for the master branch, I will go ahead and merge it.

PR #166, opened 2026-03-16
tdx-reference-ramdisk-image: fix undefined variable case

michaelopdenacker:
0e08b8d3ed: tdx-reference-ramdisk-image.bb: fix do_rootfs dependency definition introduced a regression when the DM_VERITY_IMAGE_TYPE variable is not defined, typically when dm-verity is not enabled.

In this case, the statement:
@d.getVar('DM_VERITY_IMAGE_TYPE').replace('-', '_') errors with this message:
AttributeError: 'NoneType' object has no attribute 'replace'

tdx-reference-ramdisk-image.bb is only relevant when the "tdxref-signed" class is inherited, but it is parsed in any case.

Add "or ''" to the expression, to make sure the result is always a string, on which the replace() operator can be applied.

Tested in both cases: DM_VERITY_IMAGE_TYPE defined and not defined.

### DIFF
--- a/recipes-images/images/tdx-reference-ramdisk-image.bb
+++ b/recipes-images/images/tdx-reference-ramdisk-image.bb
@@ -45,7 +45,7 @@ DEPENDS:remove = "\
 
 # rootfs should be built before the ramdisk so we have
 # dm-verity.env to add to the ramdisk
-do_rootfs[depends] += "${DM_VERITY_IMAGE}:do_image_${@d.getVar('DM_VERITY_IMAGE_TYPE').replace('-', '_')}"
+do_rootfs[depends] += "${DM_VERITY_IMAGE}:do_image_${@ (d.getVar('DM_VERITY_IMAGE_TYPE') or '').replace('-', '_')}"
 
 # ensure dm-verity.env is updated also when rebuilding DM_VERITY_IMAGE
 do_image[nostamp] = "1"


recipes-images/images/tdx-reference-ramdisk-image.bb
```diff
@@ -45,7 +45,7 @@ DEPENDS:remove = "\
 
 # rootfs should be built before the ramdisk so we have
 # dm-verity.env to add to the ramdisk
-do_rootfs[depends] += "${DM_VERITY_IMAGE}:do_image_${@d.getVar('DM_VERITY_IMAGE_TYPE').replace('-', '_')}"
+do_rootfs[depends] += "${DM_VERITY_IMAGE}:do_image_${@ (d.getVar('DM_VERITY_IMAGE_TYPE') or '').replace('-', '_')}"
```
sergioprado, line 48: @michaelopdenacker  I see the concern if `DM_VERITY_IMAGE_TYPE` is not defined, but I am not sure I understand in which situation that would actually happen.

`tdx-signed-dmverity.bbclass` already provides a default value for this variable, so hitting this case would seem to require explicitly unsetting it. That does not look like a meaningful configuration to me, since the purpose of this ramdisk image is precisely to mount the `dm-verity` image, which requires `DM_VERITY_IMAGE_TYPE` to be defined.

In addition, even with the proposed change, an empty value may still cause a failure later on, as it will set a dependency on an non-existent task:

```
do_rootfs[depends] += "tdx-reference-minimal-image:do_image_"
```

Could you clarify the scenario where you expect `DM_VERITY_IMAGE_TYPE` to be empty?

michaelopdenacker:
The problem happens with a "development" image in which I'm just inheriting "tdx-signed" (to share the same boot flow as with my "production" image), and in this case DM_VERITY_IMAGE_TYPE is indeed left undefined.

It's true "do_rootfs[depends]" refers to a non-existing task (even before my changes!) in the reference-ramdisk-image.bb recipe, but we don't care as this recipe is not used. However, it's still parsed anyway.

sergioprado:
Hmmm, so the issue is triggered at parsing time. Makes sense, as `do_rootfs[depends]` is required during parsing.

I quick experiment shows that this does not fail:

```
VAR2 = "${@d.getVar('VAR1').replace('-', '_')}"
```

But it fails when using `:=` to force expading the variable at parsing time:

```
VAR2 := "${@d.getVar('VAR1').replace('-', '_')}"
```

Alright, thanks for quickly providing the fix. I am merging this now.

michaelopdenacker:
Thanks a lot @sergioprado

PR #170, opened 2026-04-20
tdx-encrypted: add configurable dm-crypt cipher preset

sergioprado:
Add a new TDX_ENC_CIPHER variable to select the cipher algorithm and IV mode used by dm-crypt. Two presets are supported:

- aes-cbc: AES-CBC with plain IV (default, backward compatible)
- aes-xts: AES-XTS with plain64 IV (recommended for new deployments)

AES-XTS (IEEE 1619) is the industry-standard mode for storage encryption and avoids the watermarking vulnerability of CBC with a plain sector-number IV. The default is kept as aes-cbc to preserve compatibility with existing encrypted devices.

The cipher preset is validated at build time via a BitBake sanity check, and the key size is derived automatically from the preset (32 bytes for aes-cbc, 64 bytes for aes-xts).

Also enable CONFIG_CRYPTO_AES, CONFIG_CRYPTO_CBC and CONFIG_CRYPTO_XTS in the kernel to make the crypto dependencies explicit and ensure both CTS and XTS modes are available.


jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

PR #173, opened 2026-05-01
tdx-enc-handler: fix wrapped key creation on read-only rootfs

sergioprado:
When `TDX_ENC_KEY_LOCATION="partition"` is used on a system with a read-only rootfs (for example, with `dm-verity`), creating a new wrapped key fails with:

```
caam: ERROR: Error saving key blob!
```

This is a regression introduced by commit 79002ecf9229 ("tdx-enc-handler: write key blob atomically"), and happens because mkdir/mktemp cannot write to `TDX_ENC_KEY_DIR` (default `/var/local/private/.keys`).

Fix this by always setting both `TDX_ENC_KEY_DIR` and `TDX_ENC_KEY_FULLPATH` to `/tmp` when `TDX_ENC_KEY_LOCATION="partition"`.

Tested with both `TDX_ENC_KEY_LOCATION="partition"` and `TDX_ENC_KEY_LOCATION="filesystem"`, to make sure there are no further regressions.

### DIFF
--- a/recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
+++ b/recipes-core/tdx-enc-handler/tdx-enc-handler/tdx-enc.sh
@@ -128,6 +128,8 @@ tdx_enc_prepare_generic() {
     tdx_enc_log "Reserved blocks: $TDX_ENC_STORAGE_RESERVE..."
 
     if [ "${TDX_ENC_KEY_LOCATION}" = "partition" ]; then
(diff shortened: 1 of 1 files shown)


sergioprado:
@rborn-tx @jsrc27 I’m not sure whether you’re working today. Since this is a simple fix and I’d like it to make it into the next nightly build, I’m going to go ahead and merge it. If you have any concerns, please let me know.

PR #174, opened 2026-05-04
u-boot-distro-boot-harden: fix bug with root argument amendment

rborn-tx:
Add missing argument to a grep invocation which was introduced by the commit titled "u-boot-distro-boot-harden: make individual amendments configurable".

Related-to: TOR-4240


sergioprado (approved):
LGTM

sergioprado:
@jsrc27 I will move on and merge this before the next nightly runs. If you have any concerns, please let me know.

PR #176, opened 2026-05-13
Introduce Secure Debug feature

sergioprado:
This PR introduces initial Secure Debug support in `meta-toradex-security`.

The feature allows supported SoCs to restrict JTAG access through fuse programming. The initial implementation supports Verdin iMX8MM using the NXP System JTAG Controller (SJC), with support for authenticated JTAG and debug disable modes.

When enabled, the layer appends an ordered set of fuse-programming commands to `fuse-cmds.txt` and records the same fuses in `imx-config.fuse`.

Tested on Verdin iMX8MM with Lauterbach TRACE32.


classes/include/secure-debug/tdx-secure-debug-sjc.inc
```diff
@@ -0,0 +1,78 @@
...
+addhandler validate_secure_debug_sjc
+validate_secure_debug_sjc[eventmask] = "bb.event.SanityCheck"
+python validate_secure_debug_sjc() {
+    if e.data.getVar('TDX_SECURE_DEBUG_ENABLE') != '1':
+        return
+
+    machine = e.data.getVar('MACHINE')
+    supported = (e.data.getVar('TDX_SECURE_DEBUG_SUPPORTED_MACHINES') or '').split()
+    if machine not in supported:
+        return
```
jsrc27, line 52: In theory this check is redundant correct? Since the supported machine error in the bbclass would trigger and stop the build anyway.

docs/README-secure-debug.md
```diff
@@ -0,0 +1,189 @@
...
+| `TDX_SECURE_DEBUG_ENABLE` | Enable or disable the Secure Debug feature. Allowed values: `0` or `1`. | `1` |
+| `TDX_SECURE_DEBUG_MODE` | Debug policy. Allowed values: `authenticated` or `disabled`. | `authenticated` |
+
+For SJC-based SoCs (e.g. iMX8M), the following additional variables are available:
+
+| Variable | Description | Default value |
+| :------- | :---------- | :------------ |
+| `TDX_SECURE_DEBUG_KEY_FILE` | Path to the file containing the SJC response key, used when `TDX_SECURE_DEBUG_MODE = "authenticated"`. | `${TOPDIR}/keys/secure-debug/key.txt` |
+| `TDX_SECURE_DEBUG_SJC_HEO` | Block the HAB software path that can reopen JTAG from trusted boot software. Allowed values: `0` or `1`. | `1` |
+| `TDX_SECURE_DEBUG_SJC_DISABLE` | Fully disable the JTAG interface. When set to `1`, this overrides `TDX_SECURE_DEBUG_MODE` and selects the strongest available full-disable setting for the SoC. Allowed values: `0` or `1`. | `0` |
```
jsrc27, line 66: Minor feedback. On first glance the variables here were a bit confusing to understand.

Specifically:

* `TDX_SECURE_DEBUG_ENABLE` has "ENABLE" in the name and takes a boolean value
* `TDX_SECURE_DEBUG_MODE` meanwhile this has "disabled" as one of the allowed values
* `TDX_SECURE_DEBUG_SJC_DISABLE` has "DISABLE" in the name and takes boolean values

The enable/disable wording is a bit overloaded on first impression. Not sure what would be a better suggestion that is easier for general comprehension. Or, we just leave it as is which leads users to carefully read the doc.

jsrc27 (commented):
Some minor feedback.

classes/include/secure-debug/tdx-secure-debug-sjc.inc
sergioprado: Actually this is not redundant. This is here to avoid doing sanity checks on the SJC-related backend variables for machines that are not currently supported.

If one enables secure debug on `verdin-imx8mp` for example, it is expected to fail with a `machine not supported` message in the sanity check from the main class. If you remove the check above, it might fail earlier with a sanity check from the include file, which does not make sense for an unsupported machine.

docs/README-secure-debug.md
sergioprado: That is good feedback, and I am open to suggestions. For now, this is the best naming scheme I could come up with.

> `TDX_SECURE_DEBUG_ENABLE` has “ENABLE” in the name and takes a boolean value

This follows the existing convention in the layer. We already have several variables with `ENABLE` in the name that take a boolean value to turn a specific feature on or off, such as `TDX_IMX_HAB_ENABLE`, `TDX_OPTEE_ENABLE`, and `TDX_UBOOT_HARDENING_ENABLE`.

> `TDX_SECURE_DEBUG_MODE` has `disabled` as one of the allowed values

Here, `disabled` means “JTAG disabled” as the selected debug policy. I considered using `jtag-disabled`, but it felt a bit redundant in the context of a Secure Debug mode variable. But we could re-consider it.

> `TDX_SECURE_DEBUG_SJC_DISABLE` has “DISABLE” in the name and takes a boolean value

For this one, I think `DISABLE` makes sense because it controls whether we disable a specific hardware block/feature in the SJC backend.

The idea is:

```text
TDX_SECURE_DEBUG_ENABLE = "1"
```

enables the Secure Debug feature.

Then:

```text
TDX_SECURE_DEBUG_MODE = "authenticated"
```

selects authenticated JTAG as the debug policy, while:

```text
TDX_SECURE_DEBUG_MODE = "disabled"
```

selects a policy where JTAG/debug access is disabled.

Finally, on SJC-based devices, there is the lower-level override:

```text
TDX_SECURE_DEBUG_SJC_DISABLE = "1"
```

which fully disables the SJC/JTAG path, bypassing the generic mode selection.

I agree the naming is not perfect. Naming is always hard, and this is probably the right time to change it if we want to. I am open to suggestions.

recipes-bsp/imx-fuses/files/create_fuse_cmds.sh
```diff
@@ -27,13 +34,117 @@ WARNING2="\
...
+                fi
+                ;;
+            "disabled")
+                secure_debug_emit JTAG_SMODE_NODEBUG "JTAG_SMODE = No Debug"
+                ;;
+            *)
+                echo "Error: invalid TDX_SECURE_DEBUG_MODE='${TDX_SECURE_DEBUG_MODE}'" >&2
+                return 1
+                ;;
+        esac
```
rborn-tx, line 141: ### `JTAG_HEO` is skipped in `disabled` mode, leaving the HAB reopen path available

`JTAG_HEO` is emitted only in the `authenticated` arm, and `tdx-secure-debug-sjc.inc:12-18` states the same restriction ("Relevant only when `TDX_SECURE_DEBUG_MODE = "authenticated"`"), so this is deliberate rather than an oversight — but the premise looks wrong.

`JTAG_HEO` blocks the `HAB_JDE` path, and NXP's description of that path (AN4686, *Configuring Secure JTAG for the i.MX 6 Series*) is that writing `HAB_JDE` "opens the JTAG **regardless of its security mode**, without activating the challenge-response mechanism". It is not scoped to Secure JTAG mode; it is precisely the escape hatch from `No Debug` mode. So in `disabled` mode, even with the default `TDX_SECURE_DEBUG_SJC_HEO = "1"`, no `JTAG_HEO` fuse is written and trusted boot software can still re-enable JTAG.

The user-visible result is inverted: the mode documented as the stronger lock-down (`docs/README-secure-debug.md:31`) ends up strictly weaker against the HAB reopen path than `authenticated`. It is reachable with a documented, supported configuration — `TDX_SECURE_DEBUG_MODE = "disabled"` is one of the two examples at `docs/README-secure-debug.md:76-82` — and the outcome is permanent once the fuses are burnt and the device is closed.

Verified by running the generator: with `TDX_SECURE_DEBUG_MODE=disabled TDX_SECURE_DEBUG_SJC_HEO=1` the emitted section is `JTAG_SMODE = No Debug`, `KTE`, `BOOT_CFG_LOCK_OP` — no `JTAG_HEO` line.

Suggested fix — hoist the emission out of the `authenticated` arm so it applies to every non-open policy:

```sh
        "disabled")
            secure_debug_emit JTAG_SMODE_NODEBUG "JTAG_SMODE = No Debug"
            ;;
        ...
    esac
    if [ "${TDX_SECURE_DEBUG_SJC_HEO}" = "1" ]; then
        secure_debug_emit JTAG_HEO "JTAG_HEO = 1 (block HAB software reopen)"
    fi
    secure_debug_emit KTE "KTE = 1 (gate bus tracing on SJC state)"
```

and drop the "Relevant only when …" sentence from the `.inc` comment.

The same question applies to the `TDX_SECURE_DEBUG_SJC_DISABLE = "1"` branch, which also emits neither `JTAG_HEO` nor `KTE`. If the reasoning is that a disabled SJC cannot be reopened by HAB at all, that is worth one sentence in `docs/README-secure-debug.md` rather than being left implicit.

If you have already tested `disabled` mode against a HAB debug-enable attempt on hardware, that would settle this faster than any code change — I have no Verdin iMX8MM here, so this is an argument from the documented `HAB_JDE` semantics, not from an observed reopen on silicon.

recipes-bsp/imx-fuses/files/create_fuse_cmds.sh
```diff
@@ -27,13 +34,117 @@ WARNING2="\
...
+
+    declare -gA SJC_TEMPLATE
+    local prefix name bank word mask
+    while IFS=: read -r prefix name bank word mask; do
+        [ "$prefix" = "SJC" ] || continue
+        SJC_TEMPLATE[$name]="$bank $word $mask"
+    done < "$template_file"
+}
+
+secure_debug_append() {
```
rborn-tx, line 101: ### Partial output files are left behind when the secure-debug section fails

The SRK loop is careful to clean up on failure (`rm -rf "$FUSE_INFO_FILE" "$FUSE_CMDS_FILE"` at lines 167 and 178). `secure_debug_append` is not: every `secure_debug_emit` appends to both files as it goes, so a failure part-way through leaves a truncated `fuse-cmds.txt` and an `imx-config.fuse` carrying a partial `SJC:` block.

Verified: with `TDX_SECURE_DEBUG_KEY_FILE` pointing at a nonexistent path the script exits 1 and both files remain on disk, containing the SRK section plus the secure-debug header.

Impact is limited because the BitBake task fails and both `u-boot-hab.inc:252` and `imx-boot-hab.inc:202` only deploy on the success path, and the next run truncates both files before writing. Still, a half-written fuse script sitting in `WORKDIR` is exactly the kind of file someone eventually reads by hand.

Suggested fix: mirror the existing cleanup — trap the failure in `create_fuse_cmds` and remove both files before returning non-zero.

Minor, same area: a missing key file surfaces as a raw bash redirection error (`create_fuse_cmds.sh: line 119: /path/key.txt: No such file or directory`) rather than the script's own diagnostic. The class-level sanity check catches this earlier in a normal build, so it only shows when the script is run directly.

docs/README-secure-debug.md
```diff
@@ -0,0 +1,189 @@
...
+- do not commit it to the repository;
+- do not expose it in build logs;
+- store it with appropriate access control;
+- back it up securely before provisioning devices.
+
+The response key is programmed into OTP fuses. Once programmed and locked, it cannot be changed or recovered by software. If the key is lost, authenticated debug access to that device is lost as well.
+
+This build-time flow uses the same response key for all devices programmed from the same build. If that key leaks, every device provisioned with it can be opened through authenticated JTAG.
+
+If your threat model requires per-device debug credentials, provision the response key in your manufacturing flow instead of using this class to generate the response-key fuse commands.
```
rborn-tx, line 118: ### The response key lands in `DEPLOY_DIR_IMAGE` in plaintext, and this section does not say so

By construction the 56-bit response key appears verbatim in the generated `fuse-cmds.txt` and `imx-config.fuse`, and both are installed into `DEPLOY_DIR_IMAGE` (`recipes-bsp/u-boot/u-boot-hab.inc:252-255`). That is unavoidable — the fuse commands *are* the key — but the advice here covers repositories and build logs and stops there, which leaves the most likely leak path (an archived or published deploy directory, a CI artifact bundle) unmentioned.

I checked the one place that could have made this worse and it is fine: `classes/include/signed/tdx-signed-imx-hab-tezi.inc:8` filters on `^H:F:`, so the optional `prog_secure_boot_fuses` U-Boot environment command picks up only the SRK-hash lines and does not embed the response key into the Tezi image. Worth noting in this same section that this convenience command deliberately does not program the secure-debug fuses.

Suggested fix: add a bullet stating that `fuse-cmds.txt` and `imx-config.fuse` in the deploy directory contain the key in clear text and must be handled accordingly. Optionally mention that running the build with `XTRACE=1` traces the key into the task log (`create_fuse_cmds.sh:3`, `:119`).

recipes-bsp/imx-fuses/files/README-template.txt
```diff
@@ -0,0 +1,66 @@
...
+
+This directory contains template files consumed by create_fuse_cmds.sh to
+generate the per-board fuse-programming command file (fuse-cmds.txt) and the
+human-readable fuse map (imx-config.fuse).
+
+There are two template flavors:
+
+  *-template.fuse           HAB/AHAB SRK-hash + boot-close template
+                            (one per SoC family)
+
```
rborn-tx, line 12: ### `SJC` is expanded two different ways

Here it is "Secure JTAG Controller"; `classes/include/secure-debug/tdx-secure-debug-sjc.inc:1`, `create_fuse_cmds.sh:44` and `docs/README-secure-debug.md:25` all say "System JTAG Controller", which is NXP's name. Worth fixing here.

rborn-tx (commented):
Review by Claude for assessment/discussion.

---

Besides the inline comments:

### Pre-existing `return` outside a function on the invalid-SOC path

`recipes-bsp/imx-fuses/files/create_fuse_cmds.sh:231-234`

Not introduced here, but the file is being touched. `return 1` at top level makes bash print `return: can only 'return' from a function or sourced script`; the script does still exit non-zero (status 2), so the behaviour is correct by accident. Verified by running the script with `SOC=BOGUS`. `exit 1` would be the honest spelling.

sergioprado:
@rborn-tx Did you validate Claude's output before posting it here? Do you think all of the generated comments are valid and relevant?

If so, could you please add your own assessment to each of Claude's comments to confirm that you agree with them?

I'm perfectly fine with using AI for code reviews — I use it myself. However, I think a human should remain part of the review process and filter out invalid or low-value comments before posting them. Otherwise, it can add unnecessary noise to the review and waste development time.

Thanks!

rborn-tx:
> @rborn-tx Did you validate Claude's output before posting it here? Do you think all of the generated comments are valid and relevant?
> 
> If so, could you please add your own assessment to each of Claude's comments to confirm that you agree with them?
> 
> I'm perfectly fine with using AI for code reviews — I use it myself. However, I think a human should remain part of the review process and filter out invalid or low-value comments before posting them. Otherwise, it can add unnecessary noise to the review and waste development time.
> 
> Thanks!

Claude found more items which I cleared myself before posting. The ones I posted are those that either I consider sensible (on a quick pass) or that would required a more detailed investigation (like the `JTAG_HEO is skipped in disabled mode, leaving the HAB reopen path available`).

As I wrote in the initial comment of the thread, the idea of posting an AI review is to assess/discuss. Of course, we don't necessarily need to accept any of them. However, having the record of possible issues as part of the PR looks useful to me, even if we decide to dismiss some/most of them.

Anyways, that's something we can talk about. Lately, given the increased number and size of PRs/MRs I've started to do a pass with AI -> quick filtering -> posting, before entering more detailed/time consuming passes as a way to optimize work. But I'd be okay to have a different workflow for this repository :wink:.

docs/README-secure-debug.md
jsrc27: After some thought here's my opinion:

* `TDX_SECURE_DEBUG_ENABLE`: This can stay as is, it's consistent with other `*_ENABLE` variables across the meta-layer as you've said.
* `TDX_SECURE_DEBUG_MODE`: Maybe we can think of another word for the `disabled` mode. Limited? Restricted? It's just a bit off-putting that this has a mode called `disabled` when the previous variable is about enabling/disabling. Which I know they refer to enabling/disabling different things, but on quick glance of these variables it can be confusing when talking about enabling/disabling in a broader context.
* `TDX_SECURE_DEBUG_SJC_DISABLE`: This one is a bit tough to figure out an alternative for. It overrides the previous variable so it would be nice if the name clearly connected to this fact. That way the variable behavior is more apparent just based on the name. Maybe something like `TDX_SECURE_DEBUG_SJC_MODE`, or `TDX_SECURE_DEBUG_SJC_MODE_OVERRIDE`. Hmm not sure if these are universally clear though.

sergioprado:
@rborn-tx Alright, I will review those comments.

For next time, though, I would kindly ask you not to simply copy and paste the tool’s output. It is perfectly fine to mention that the code was reviewed by tool XXXX, but please write the comment in your own words as well. That makes it clear that you reviewed and understood the issue being reported, instead of just forwarding the tool’s review.

docs/README-secure-debug.md
sergioprado: > TDX_SECURE_DEBUG_ENABLE: This can stay as is, it's consistent with other *_ENABLE variables across the meta-layer as you've said.

Right. Will leave it as-is.

> TDX_SECURE_DEBUG_MODE: Maybe we can think of another word for the disabled mode. Limited? Restricted? It's just a bit off-putting that this has a mode called disabled when the previous variable is about enabling/disabling. Which I know they refer to enabling/disabling different things, but on quick glance of these variables it can be confusing when talking about enabling/disabling in a broader context.

I changed it to `no-debug` instead of using "limited" or "restricted", because "No Debug" is NXP's own name for that state. So `TDX_SECURE_DEBUG_MODE` now accepts `authenticated` or `no-debug`. A user who sees `no-debug` in our documentation will find the same term in the SoC Security Reference Manual.

> TDX_SECURE_DEBUG_SJC_DISABLE: This one is a bit tough to figure out an alternative for. It overrides the previous variable so it would be nice if the name clearly connected to this fact. That way the variable behavior is more apparent just based on the name. Maybe something like TDX_SECURE_DEBUG_SJC_MODE, or TDX_SECURE_DEBUG_SJC_MODE_OVERRIDE. Hmm not sure if these are universally clear though.

For `TDX_SECURE_DEBUG_SJC_DISABLE`, I would keep the current name. `SJC_MODE` would be misleading because the variable is boolean, not a mode, and it would sit next to the actual mode variable. `SJC_MODE_OVERRIDE` describes the mechanism rather than the effect. The current name identifies a specific hardware control inside the backend-specific namespace, which seems appropriate for a backend variable. The override behavior is documented in the table where the user first sees the variable.

recipes-bsp/imx-fuses/files/create_fuse_cmds.sh
sergioprado: I think the AN4686 reference may be incorrect here. I could not find the quoted wording in AN4686 Rev. 4.0. The phrase “regardless of its security mode” does not appear in that document, and the cited title, “Configuring Secure JTAG for the i.MX 6 Series,” also does not match AN4686, whose title is “Secure Debug in i.MX 6/7/8M Family of Applications Processors.”

From both AN4686 and the Security Reference Manual, the scope looks clear to me: `JTAG_HEO` is only meaningful in authenticated/Secure JTAG mode.

- AN4686 §4.3 frames `HAB_JDE` as a way to bypass Secure JTAG authentication: "The Secure JTAG authentication can be bypassed in SW, by writing 1 to the HAB_JDE".
- The Security Reference Manual is also explicit in Table 7-10: `JTAG_HEO` "Disables the HAB Unlock JTAG CSF command intended for JTAG_SMODE = Secure."

If `HAB_JDE` only applies in Secure JTAG mode, then programming `JTAG_HEO` in `no-debug` mode would burn an OTP fuse that has no effect. Worse, it could lead users to program an irreversible fuse unnecessarily.

If the application note you are referring to says something different, please point me to it and reopen this therad so we can review it together.

That said, it makes sense to document that `TDX_SECURE_DEBUG_SJC_HEO` only applies to authenticated mode. I made that explicit in `README-secure-debug.md`.

docs/README-secure-debug.md
sergioprado: Yes, this is true and by design.

The key is present in `fuse-cmds.txt` and `imx-config.fuse` by construction, and both files are installed into the deploy directory by `u-boot-hab.inc` for HAB and `imx-boot-hab.inc` for AHAB. This is unavoidable because the fuse commands are the key material that will be programmed into OTP fuses.

I added a note about this in the documentation.

recipes-bsp/imx-fuses/files/README-template.txt
sergioprado: "System JTAG Controller" is NXP's terminology, as used in AN4686, so `README-template.txt` was wrong indeed. Fixed.

recipes-bsp/imx-fuses/files/create_fuse_cmds.sh
sergioprado: I think this is a case where the tool is being noisy and overreacting to the code structure.

None of the `secure_debug_emit` calls inside `secure_debug_append` check the return value explicitly. That is intentional: `set -e` makes a failing emit abort the script. If we wrap each call just to catch the failure, `set -e` is suspended for the whole call, including inside the function.

In practice, the BitBake task fails, and I do not think partially generated files in `WORKDIR` are a problem, since they will be trucated in the next run.

> Minor, same area: a missing key file surfaces as a raw bash redirection error (create_fuse_cmds.sh: line 119: /path/key.txt: No such file or directory) rather than the script's own diagnostic. The class-level sanity check catches this earlier in a normal build, so it only shows when the script is run directly.

That concern is valid only if the script is called directly, which is not the intended use case here. In a normal build, the class sanity check catches this earlier and fails with an explicit error message.

sergioprado:
@rborn-tx @jsrc27 This is ready for another review.

docs/README-secure-debug.md
jsrc27: > I changed it to no-debug instead of using "limited" or "restricted", because "No Debug" is NXP's own name for that state. So TDX_SECURE_DEBUG_MODE now accepts authenticated or no-debug. A user who sees no-debug in our documentation will find the same term in the SoC Security Reference Manual.

I like this change, removes another enable/disable keyword from this set of variables and it correlates back to NXP documentation for easier cross-referencing.

> For TDX_SECURE_DEBUG_SJC_DISABLE, I would keep the current name. SJC_MODE would be misleading because the variable is boolean, not a mode, and it would sit next to the actual mode variable. SJC_MODE_OVERRIDE describes the mechanism rather than the effect. The current name identifies a specific hardware control inside the backend-specific namespace, which seems appropriate for a backend variable. The override behavior is documented in the table where the user first sees the variable.

This is acceptable for me.

jsrc27 (approved):
LGTM

docs/README-secure-debug.md
rborn-tx, line 1: @sergioprado It looks like this file is not referenced from anywhere. Did you intend to refer to it from the top-level README.md?

docs/README-secure-debug.md
```diff
@@ -0,0 +1,193 @@
...
+## Secure Debug on iMX8M
+
+On iMX8M, Secure Debug uses the NXP **System JTAG Controller (SJC)**. The SJC can gate the JTAG interface and require a challenge/response authentication flow before allowing debug access.
+
+The supported operating modes are:
+
+- **JTAG Enable**: JTAG fully open (default state). Used during development.
+- **Authenticated JTAG**: JTAG access is blocked after reset and can only be reopened by a debugger that knows the programmed response key.
+- **No Debug**: security-sensitive debug features, such as CPU halt and memory access, are disabled. Some lower-risk JTAG features, such as boundary scan, may still remain available depending on the SoC.
+- **Full JTAG Disable**: the JTAG interface is fully disabled. Even boundary scan is unavailable. This is controlled through an SJC-specific variable.
```
rborn-tx: Is the terminology here supposed to come from NXP documentation? I'm asking because checking [AN4686: Secure Debug in i.MX 6/7/8M Family of Applications Processors](https://docs.nxp.com/bundle/AN4686/page/topics/introduction.html) it says there are 3 security modes (plus one where the SJC is disabled):

- Mode 1: No Debug (same term as NXP)
- Mode 2: Secure JTAG (maps to our "Authenticated JTAG")
- Mode 3: JTAG Enabled (maps to our "JTAG Enable")
- option to disable the SJC functionality: no Mode specified (maps to our "Full JTAG Disable")

**If we wanted to align with NXP** then we'd have something like this, from most secure to least secure:

- JTAG Disabled
- No Debug
- Secure JTAG
- JTAG Enabled

Unless you've used another source of information from NXP, I think aligning the terminology is helpful to those checking what this layer does against the SoC documentation.

By the way, wouldn't to good to have pointers to the Application Notes or other NXP material you've used as references for the implementation?

docs/README-secure-debug.md
```diff
@@ -0,0 +1,193 @@
...
+
+Besides the mode itself, the backend also configures whether the HAB/AHAB software path can reopen JTAG from trusted boot code, whether the trace port is gated by the debug policy, and the lock bits that prevent the whole configuration from being overridden afterwards.
```
rborn-tx: Since this section is about the iMX8M, maybe we shouldn't mention AHAB here?

rborn-tx (commented):
(Human-pass 😉) Overall looks great.

Raising a few questions about minor details...

docs/README-secure-debug.md
sergioprado: > (Human-pass 😉) Overall looks great.

Nice talking to humans! :-)

> @sergioprado It looks like this file is not referenced from anywhere. Did you intend to refer to it from the top-level README.md?

Good catch! I completly forgot to update the main `README` file. I added a small entry in the "supported features" session and included a link to the new document.

docs/README-secure-debug.md
sergioprado: Hmm, it would be nice to have a common terminology, but unfortunately the terminology varies across vendors and SoC families.

Since the plan is to support Secure Debug across different SoC families, such as i.MX6/7/8, i.MX9 with ELE, and TI K3, I tried to define a terminology that is easy to understand and reasonably consistent across all of them.

> By the way, wouldn't to good to have pointers to the Application Notes or other NXP material you've used as references for the implementation?

Good idea. I added a session in the end referencing the documents I used during the implementation.

docs/README-secure-debug.md
sergioprado: Good catch. Removed.

sergioprado:
@rborn-tx This is ready for another review.

docs/README-secure-debug.md
rborn-tx: > Hmm, it would be nice to have a common terminology, but unfortunately the terminology varies across vendors and SoC families.

Just to clarify, here I'm just pointing out that the terminology used in the README, **in the "Secure Debug on iMX8M" section** does not match the **iMX8M manual/application-note** terminology of that SoC family (those present in their AN4686).

sergioprado: Hmm, I missed your point before. You are right; it makes total sense to align this with the NXP documentation.

I changed the documentation to describe only the three operating modes: **No Debug**, **Secure JTAG**, and **JTAG Enabled**. I also added a separate paragraph mentioning the option to fully disable the SJC.

The PR has been updated.

rborn-tx: Nice!

PR #177, opened 2026-05-26
Add HSM-backed signing support for TI K3 boot containers

sergioprado:
This PR extends HSM-backed signing to support the TI K3 boot container, completing HSM coverage across the secure boot signing chain for TI K3 targets.

On TI K3, the boot container is signed during the U-Boot build by U-Boot's `binman` tool through OpenSSL-driven `x509_cert` entries. To allow a PKCS#11 URI to be used in place of an on-disk key file, this PR adds a downstream patch to binman's `x509_cert` entry type. The patch is a minimal subset of the upstream v5 series in [1].

The implementation was build and runtime-tested on AM62 with both SoftHSM and YubiKey. It was also build and runtime-tested on i.MX8MM to verify that no regressions were introduced.

[1] https://patchwork.ozlabs.org/project/uboot/patch/20260525132828.3348679-3-sergio.prado@e-labworks.com/

### DIFF
--- a/classes/include/signed/tdx-signed-hsm.inc
+++ b/classes/include/signed/tdx-signed-hsm.inc
@@ -29,12 +29,8 @@ UBOOT_MKIMAGE_SIGN_ARGS:append = " -N pkcs11"
 # SoftHSM configuration file
 TDX_SIGNED_HSM_PKCS11_SOFTHSM_CONF ?= "${TOPDIR}/keys/softhsm/conf/softhsm2.conf"
 
-# HSM-backed HAB signing requires a CST binary with PKCS#11 support.
-# Since the externally provided CST may not include that support, force
-# the use of a source-built CST and enable its PKCS#11 integration when
-# HAB signing is active.
-TDX_IMX_HAB_CST_BUILD_FROM_SOURCE = "${TDX_IMX_HAB_ENABLE}"
-TDX_IMX_HAB_CST_BUILD_WITH_PKCS11 = "${TDX_IMX_HAB_CST_BUILD_FROM_SOURCE}"
+# NXP imx-specific HSM configuration
+require ${@ 'tdx-signed-hsm-imx.inc' if 'imx-generic-bsp' in d.getVar('MACHINEOVERRIDES_EXTENDER').split(':') else ''}
 
 # suppress warning messages from this class
 TDX_SIGNED_HSM_SUPPRESS_WARNINGS ?= "0"
--- a/classes/include/signed/tdx-signed-hsm.inc
+++ b/classes/include/signed/tdx-signed-hsm.inc
@@ -22,8 +22,8 @@ TDX_SIGNED_HSM_FIT_TOKEN_LABEL ?= ""
 
 # FIT image configuration to use the HSM for signing
 FIT_GENERATE_KEYS = "0"
-UBOOT_SIGN_KEYDIR = "${TDX_SIGNED_HSM_FIT_TOKEN_URL}"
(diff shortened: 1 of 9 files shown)


jsrc27 (approved):
LGTM

classes/include/signed/tdx-signed-hsm.inc
```diff
@@ -22,19 +22,19 @@ TDX_SIGNED_HSM_FIT_TOKEN_LABEL ?= ""
 
 # FIT image configuration to use the HSM for signing
 FIT_GENERATE_KEYS = "0"
-UBOOT_SIGN_KEYDIR = "${TDX_SIGNED_HSM_FIT_TOKEN_URL}"
-UBOOT_SIGN_KEYNAME = "${TDX_SIGNED_HSM_FIT_TOKEN_LABEL}"
+UBOOT_SIGN_KEYDIR:forcevariable = "${TDX_SIGNED_HSM_FIT_TOKEN_URL}"
+UBOOT_SIGN_KEYNAME:forcevariable = "${TDX_SIGNED_HSM_FIT_TOKEN_LABEL}"
```
rborn-tx, line 26: Would it be possible to avoid this use of override `forcevariable` here? Maybe use something less strong such as `tdx-signed` to leave the `forcevariable` as a last resort method for a user of the layer.

recipes-bsp/u-boot/files/0001-binman-x509_cert-support-PKCS-11-URI-in-keyfile-for-.patch
```diff
@@ -0,0 +1,112 @@
...
+non-interactively. The preferred way to deliver the PIN is via
+pkcs11-module-token-pin in openssl.cnf.
+
+This is the minimal subset of the upstream v5 series sent on
+2026-05-25; documentation, tests, and the test fixture are omitted
+to minimise rebase friction when U-Boot is bumped.
+
+Upstream-Status: Submitted [https://lists.denx.de/pipermail/u-boot/2026-May/619780.html]
+
+Signed-off-by: Sergio Prado <sergio.prado@e-labworks.com>
```
rborn-tx, line 26: As a comment, and I don't know if the upstream project cares about that, but we are now trying to make patches compatible with [git trailers](https://git-scm.com/docs/git-interpret-trailers) like the BSP team does. That means there would be no empty line between tags.

rborn-tx (commented):
Looks great!

Raised a couple of questions before approving.

classes/include/signed/tdx-signed-hsm.inc
sergioprado: Yeah, this was required because TI machines set these variables using an override.

I would also prefer to avoid `forcevariable` here, but I thought about it and decided it is probably the better long-term solution for two reasons:

- Even if we find a weaker override that works today, we may run into the same problem again when adding support for a new machine that uses a different or stronger override.
- In this specific case, we really do want to force the value of these variables. They are not intended to be user-configurable when HSM is enabled.

That said, I am open to other approaches if you have a better suggestion.

recipes-bsp/u-boot/files/0001-binman-x509_cert-support-PKCS-11-URI-in-keyfile-for-.patch
sergioprado: Good catch! I removed the extra line. Thanks for spotting it!

classes/include/signed/tdx-signed-hsm.inc
rborn-tx: Yeah, that makes sense.

rborn-tx (approved):

PR #178, opened 2026-05-28
u-boot: allow generating bootloader suitable for SDP loading (iMX6ULL/iMX7)

rborn-tx:
Allow generating a bootloader binary suitable for SDP loading, getting rid of the HAB events that happened before on the iMX6ULL and iMX7.

Also, consider the memory copy (CP) and Fastboot (FB) protections to be "implemented" on machines based on those SoCs, now that they have been runtime tested with a build of TEZI.

Build tests have been performed with Torizon OS and TDXREF images to ensure the build still succeeds (runtime tests are left for the test system).

Related-to: TOR-4400, TOR-4431, TOR-4432

### DIFF
--- a/classes/imx-hab.bbclass
+++ b/classes/imx-hab.bbclass
@@ -142,3 +142,12 @@ TDX_IMX_HAB_CST_SGK_SUPP  ?= "1"
 
 # Only SRK is supported in the current released firmware (April 2025) for iMX9 based SoCs
 TDX_IMX_HAB_CST_SGK_SUPP:mx9-generic-bsp ?= "0"
+
+# List of U-Boot configurations targeting SDP loading; "*" means all;
+# normally this would be a subset of the values in UBOOT_CONFIG
+TDX_IMX_HAB_SDP_CONFIGS ?= ""
+
+# Internal version of the above variable making it relevant on specific SoCs
+TDX_IMX_HAB_SDP_CONFIGS_INTERNAL ?= ""
+TDX_IMX_HAB_SDP_CONFIGS_INTERNAL:mx6ull-generic-bsp = "${TDX_IMX_HAB_SDP_CONFIGS}"
+TDX_IMX_HAB_SDP_CONFIGS_INTERNAL:mx7-generic-bsp = "${TDX_IMX_HAB_SDP_CONFIGS}"
--- a/docs/README-secure-boot-imx.md
+++ b/docs/README-secure-boot-imx.md
@@ -36,18 +36,39 @@ After that, configure the various variables listed below to match your choices; 
 | `TDX_IMX_HAB_CST_DIG_ALGO` | Digest algorithm as entered into the CST tool. | `sha256` |
 | `TDX_IMX_HAB_CST_SRK_CA` | Whether or not the SRK certificates have the CA flag set as entered into the CST tool; allowed values: `0` or `1`. | `1` |
 | `TDX_IMX_HAB_CST_SRK_INDEX` | Index of the SRK to be used for signing within the SRK table; allowed values: `1`..`4`, corresponding to `SRK1`..`SRK4`, respectively. | `1` |
+| `TDX_IMX_HAB_SDP_CONFIGS` | List of U-Boot configurations targeting SDP loading; `*` means all (relevant mainly to the iMX6ULL and iMX7) | Empty |
 | `TDX_IMX_HAB_GEN_UBOOT_FUSING_CMD` | Add a function called `prog_secure_boot_fuses` to U-Boot's environment to program the secure boot fuses. Since this is intended mostly for development/tests, the function will only program the fuses but not close the device. If you enable it and run the function inside U-Boot, be aware that it will write to One-Time Programmable e-fuses, and the operation is irreversible! Allowed values are: `0` (disabled) or `1` (enabled). | `0` |
 
 The complete list of variables can be found in the `imx-hab.bbclass` file.
 
 **NOTE**: For HAB signing, [libfaketime](https://github.com/wolfcw/libfaketime) is used when generating the CSF binaries with CST in order to create reproducible bootloader image builds.
 
-### Known issues
+### Bootloaders targeting SDP loading
+
+This section is currently relevant to the iMX6ULL and iMX7 SoCs where a bootloader aiming SDP loading requires a special signing procedure.
+
+The variable `TDX_IMX_HAB_SDP_CONFIGS` can be configured to generate a bootloader suitable for SDP loading and execution.
+
+For builds where multiple U-Boot configurations are specified (e.g., `UBOOT_CONFIG = "config1 config2 config3"`), setting `TDX_IMX_HAB_SDP_CONFIGS = "*"` ensures that all configurations are built with the required modifications for SDP usage.
+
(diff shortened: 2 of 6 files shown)


docs/README-secure-boot-imx.md
```diff
@@ -36,18 +36,30 @@ After that, configure the various variables listed below to match your choices;
 | `TDX_IMX_HAB_CST_DIG_ALGO` | Digest algorithm as entered into the CST tool. | `sha256` |
 | `TDX_IMX_HAB_CST_SRK_CA` | Whether or not the SRK certificates have the CA flag set as entered into the CST tool; allowed values: `0` or `1`. | `1` |
 | `TDX_IMX_HAB_CST_SRK_INDEX` | Index of the SRK to be used for signing within the SRK table; allowed values: `1`..`4`, corresponding to `SRK1`..`SRK4`, respectively. | `1` |
+| `TDX_IMX_HAB_SDP_CONFIGS` | List of U-Boot configurations targeting SDP loading; `*` means all (relevant mainly to the iMX6ULL and iMX7) | Empty |
```
jsrc27, line 39: Just to make sure I understand the usage of this variable. If the user wants to generate the bootloader for SDP purposes, they should set this to `*`, in the simple case?

rborn-tx: Yes, that's right.

As an example, if multiple versions of U-Boot are being built (e.g. `UBOOT_CONFIG = "config1 config2 config3"`) and one sets `TDX_IMX_HAB_SDP_CONFIGS = "*"` then all of them would be built (with the amendments) for SDP usage.

More commonly though, only a subset of the configurations would aim SDP. E.g. `UBOOT_CONFIG = "nand emmc recovery"` and a user would likely set `TDX_IMX_HAB_SDP_CONFIGS = "recovery"` so only that version is built for SDP.

A special case is where `UBOOT_CONFIG` is not being used (the old, single configuration setup). In that case, the configuration doesn't have a name (type) and the user would need to set `TDX_IMX_HAB_SDP_CONFIGS = "*"`.

I made it in this way to avoid the need for having two different variables, one for dealing with the case where `UBOOT_CONFIG` is in use (name based) and another for when it is not (boolean).

sergioprado: @rborn-tx I had the same question when I was reading the code. Maybe you could add this explanation to the README file?

classes/imx-hab.bbclass
```diff
@@ -142,3 +142,12 @@ TDX_IMX_HAB_CST_SGK_SUPP  ?= "1"
 
 # Only SRK is supported in the current released firmware (April 2025) for iMX9 based SoCs
 TDX_IMX_HAB_CST_SGK_SUPP:mx9-generic-bsp ?= "0"
+
+# List of U-Boot configurations targeting SDP loading; "*" means all;
+# normally this would be a subset of the values in UBOOT_CONFIG
+TDX_IMX_HAB_SDP_CONFIGS ?= ""
```
sergioprado, line 148: Would it make sense to validate `TDX_IMX_HAB_SDP_CONFIGS` against `UBOOT_CONFIG` when `UBOOT_CONFIG` is set? A typo would currently be silently ignored, producing a normally signed image instead of an SDP-suitable one.

recipes-bsp/u-boot/u-boot-harden.inc
```diff
@@ -231,10 +231,22 @@ SRC_URI:append = "\
 "
 
 CP_PROT_SUPPORTED ?= "0"
+CP_PROT_SUPPORTED:apalis-imx6 ?= "1"
+CP_PROT_SUPPORTED:colibri-imx6 ?= "1"
+CP_PROT_SUPPORTED:colibri-imx6ull ?= "1"
+CP_PROT_SUPPORTED:colibri-imx6ull-emmc ?= "1"
+CP_PROT_SUPPORTED:colibri-imx7 ?= "1"
```
sergioprado, line 238: Both `colibri-imx6ull` and `colibri-imx7` are NAND-based variants. Currently, we do not list them as supported machines in the layer. Do we need them here?

sergioprado (changes requested):
@rborn-tx Great addition to the layer!

I left a couple of comments.

Also, could you please improve the commit message for `u-boot: allow generating bootloader suitable for SDP loading`? It would be helpful to explain what differs between a normal boot and an SDP boot with respect to signed images, and why this change is needed.

Having that context in the commit message would help others, and our future selves, better understand why the change was introduced.

docs/README-secure-boot-imx.md
rborn-tx: Sure! I'll do that in v2.

classes/imx-hab.bbclass
rborn-tx: That seems like a good DX improvement. I'll do it in v2.

recipes-bsp/u-boot/u-boot-harden.inc
rborn-tx: Unfortunately this is needed because Toradex Easy Installer is built with `MACHINE` set to the NAND variant both on the Colibri i.MX6ULL and i.MX7. It then uses different U-Boot configs (namely, `nand`, `emmc` and `recoverytezi`) to produce each bootloader variant.

By the way, the use of these overrides here is a short-term solution to stop people from trying to use the features ("cp" and "fb" protections) while they have not been tested on each specific machine. Once these features are tested, my idea was to replace the overrides by generic ones (`imx6-generic...`).

rborn-tx:
> Also, could you please improve the commit message for `u-boot: allow generating bootloader suitable for SDP loading`? It would be helpful to explain what differs between a normal boot and an SDP boot with respect to signed images, and why this change is needed.
> 
> Having that context in the commit message would help others, and our future selves, better understand why the change was introduced.

The code comments have that information but adding it to the commit message as well is sensible too. I'll do it on v2.

docs/README-secure-boot-imx.md
rborn-tx: Thanks guys for the feedback. Done in v2.

classes/imx-hab.bbclass
rborn-tx: Done in v2.

recipes-bsp/u-boot/u-boot-harden.inc
sergioprado: Gotcha.

docs/README-secure-boot-imx.md
```diff
@@ -36,18 +36,39 @@ After that, configure the various variables listed below to match your choices;
...
 The complete list of variables can be found in the `imx-hab.bbclass` file.
 
 **NOTE**: For HAB signing, [libfaketime](https://github.com/wolfcw/libfaketime) is used when generating the CSF binaries with CST in order to create reproducible bootloader image builds.
 
-### Known issues
+### Bootloaders targeting SDP loading
+
+This section is currently relevant to the iMX6ULL and iMX7 SoCs where a bootloader aiming SDP loading requires a special signing procedure.
+
+The variable `TDX_IMX_HAB_SDP_CONFIGS` can be configured to cause the generation a bootloader suitable for SDP loading and execution.
```
sergioprado: Maybe we can improve the wording here: `can be configured to cause the generation a bootloader` -> `can be configured to generate a bootloader.`

sergioprado (approved):
Looks good to me! I just left a small comment in the README.

docs/README-secure-boot-imx.md
rborn-tx: @jsrc27 Could you please take another look to see if the usage of the variable is clear now? Thanks!

jsrc27 (approved):

PR #180, opened 2026-06-09
README-secure-boot-k3.md: clarify key storage and signing chain

sergioprado:
Improve the TI K3 secure boot documentation in two areas where users have hit gaps in the past:

- Reframe the key generation example as a development-only flow and add a "A note on key storage" paragraph after the example. It warns that storing the private SMPK/BMPK on the build host exposes the root of trust to anyone with access to the machine and points users to the HSM-backed signing workflow.
- Add a new "Signed artifacts and verification chain" subsection enumerating the three signed bootloader stages (tiboot3.bin, tispl.bin, u-boot.img), explaining that each is wrapped in a ti-secure x509 certificate signed with the SMPK, and describing how the boot ROM and TIFS verify them at boot.

### DIFF
--- a/docs/README-secure-boot-k3.md
+++ b/docs/README-secure-boot-k3.md
@@ -48,7 +48,7 @@ For signature checking, the following algorithms are supported:
 - RSA (2048 and 4096)
 - ECDSA (secp256 and secp521)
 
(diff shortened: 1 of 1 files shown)


jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

PR #183, opened 2026-06-23
Skip unsupported SM-to-ELE DVFS notifications

sergioprado:
Apply a patch to prevent System Manager from sending DVFS notification messages to ELE.

This is required because the ELE firmware shipped with the current BSP, version 2.0.2.1, does not implement this API. Without this patch, ELE rejects the messages and logs AHAB events, which prevents users from closing the device with `ahab_close`.


sergioprado:
Fixed in the BSP in commit https://git.toradex.com/toradex-manifest.git/commit/?id=967aded1

PR #187, opened 2026-07-30
Report HAB/AHAB events through the hab_status/ahab_status exit code

sergioprado:
The U-Boot `hab_status` and `ahab_status` commands print the HAB/AHAB events they find but always return success, so the only way to check the result of the verification from a boot script is to parse their output.

This PR adds patches making both commands return an error whenever an event is reported, on every i.MX SoC family supported by this layer. A well-known RNG self-test event on i.MX6 SoCs is deliberately ignored and does not make the command fail.

Three of the four patches were already carried in `u-boot-fuse.inc` in the `meta-toradex-torizon` layer, which even had a TODO suggesting this move. They are brought here so that the feature is available to every user of this layer and not only to Torizon OS.

The fourth patch is new, and covers a gap. i.MX6, i.MX7 and i.MX8M implement `hab_status` in `arch/arm/mach-imx/hab.c`; i.MX8QM and i.MX8QXP implement `ahab_status` in `arch/arm/mach-imx/imx8/ahab.c` through SECO; and i.MX9x implement it in `arch/arm/mach-imx/ele_ahab.c` through the EdgeLock Enclave. The SECO file is not compiled at all on ELE based SoCs, so they needed their own patch. The right patch set is now selected per machine in the new
`u-boot-hab-status.inc`, which is included from `u-boot-hab.inc`.

Tested on Colibri iMX6, Verdin iMX8MP, Apalis iMX8 and Aquila iMX95. The Aquila iMX95 was verified both with and without ELE events, confirming that `ahab_status` only returns an error when events are present.

### DIFF
--- a/docs/README-secure-boot-imx.md
+++ b/docs/README-secure-boot-imx.md
@@ -72,7 +72,11 @@ The DCD address can be determined from the `mkimage` output logs produced by U-B
 
 If HAB/AHAB is enabled, at the end of the build, a file with the commands to fuse the SoC (`fuse-cmds.txt`) will be generated in the images directory. The commands in this file should be executed in the U-Boot command line interface.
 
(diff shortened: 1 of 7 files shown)


sergioprado:
Obs.: This should be merged together with https://github.com/torizon/meta-toradex-torizon/pull/587.

jsrc27 (approved):
LGTM

rborn-tx (approved):
LGTM

PR #188, opened 2026-08-04
Document that the encryption TPM backend always uses tpm0

sergioprado:
On systems with more than one TPM (e.g. a discrete TPM plus an OP-TEE fTPM), Linux exposes several device nodes (/dev/tpm0, /dev/tpm1, ...), which raises the question of which one manages the encryption key.

The kernel Trusted Keys subsystem is hardwired to the lowest-index chip returned by tpm_default_chip() (tpm0), and there is no build-time or runtime option to select a different one. The only way to pick a specific TPM is to control which one enumerates as tpm0 via device driver probe order, which is outside the scope of this layer.

Document this in the 'Notes on using a TPM' section so users are not misled into expecting a selectable device node.


jsrc27 (approved):
LGTM

rborn-tx (approved):

PR #190, opened 2026-08-24
Decode i.MX93 OEM secure world closed state

sergioprado:
The i.MX93 lifecycle switch in `tdx_secboot_ele_ahab_dev_is_closed()` does not handle `0x10`, which is the OEM Secure World Closed state.

Currently, this state falls through to the default case and is reported as unknown. The function still returns the correct result, but only because the default case assumes unknown states are closed. Decode this state explicitly instead.

While at it, improve the comments for both i.MX93 and i.MX95:

* OEM Locked is documented as a closed state. The Security Reference Manuals for both SoCs describe it as OEM Closed, with debug disabled and no way back through Field Return.
* Software does run in the Field Return states. The manuals list the SoC cores as active, with JTAG enabled and no authentication required for OEM containers. Reporting the device as not closed in these states is still correct, but for that reason, not because nothing runs.

Also print the lifecycle value in the Field Return message, as the neighboring cases already do.

No functional change, as every previously reachable lifecycle value is decoded to the same result as before.


jsrc27 (approved):
LGTM

rborn-tx (approved):

PR #191, opened 2026-09-01
u-boot: add switch for reproducible K3 bootloader binaries

rborn-tx:
The certificates generated when signing the K3 boot chain carry a random serial number and a validity period taken from the current time, so two builds of the same sources produce different tiboot3.bin, tispl.bin and u-boot.img. That rules out the simplest check available to anyone re-signing those binaries with their own keys: compare the result byte for byte against the one the build produced.

Here we patch binman to make the certificates deterministic (enabled by setting TDX_K3_SECBOOT_REPRODUCIBLE="1"). The variable defaults to 0 so that the change is not a breaking one within the current release branch; it is transitional and is to be removed once the behavior becomes unconditional.

The patch derives the serial number from the data being signed and runs the validity period from SOURCE_DATE_EPOCH to 20491231235959Z.  That end date looks arbitrary but is not: the K3 boot ROM cannot parse a certificate whose validity is encoded as GeneralizedTime, and X.509 requires GeneralizedTime for any date from 2050 on. A Verdin AM62 HS-FS signed with the 99991231235959Z of RFC 5280 section 4.1.2.5, the "no well-defined expiration date" convention, does not boot at all and prints nothing on the console; the same image signed with a 2049 or a 2048 date boots normally. 20491231235959Z is the latest instant expressible as UTCTime and is the workaround TI gives for the same limitation on AM64x [1]. For this reason, the constant carries a comment saying it must not be raised.

[1] https://e2e.ti.com/support/processors-group/processors/f/processors-forum/1491915/am6412-ti-supports-for-tiboot3-certificates-with-validity-beyond-2049

Related-to: TOR-4543

### DIFF
--- a/classes/include/signed/tdx-signed-k3-secboot.inc
+++ b/classes/include/signed/tdx-signed-k3-secboot.inc
@@ -11,6 +11,11 @@ TDX_K3_SECBOOT_KEY_DIR ?= "${TOPDIR}/keys/ti"
 # variable TDX_K3_SECBOOT_TARGET_HSSE_DEVICE:
 TDX_K3_SUPPRESS_HSSE_WARNINGS ?= "0"
 
+# Whether to generate the bootloader signing certificates reproducibly, so that
+# two builds of the same sources produce byte-identical tiboot3.bin, tispl.bin
+# and u-boot.img; allowed values: 0 or 1.
+TDX_K3_SECBOOT_REPRODUCIBLE ?= "0"
+
 # Choose between HSSE/HSFS variants of tiboot3.bin:
 def target_hsse(d):
     return bb.utils.to_boolean(d.getVar('TDX_K3_SECBOOT_ENABLE')) and \
--- a/recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
+++ b/recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
@@ -1,0 +1,257 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Rogerio Guerra Borin <rogerio.borin@toradex.com>
+Date: Tue, 11 Aug 2026 16:55:25 -0300
+Subject: [PATCH] binman: openssl: Make generated certificates reproducible
+
+All four certificate helpers in the openssl bintool run 'openssl req
+-new -x509' without setting a serial number or a validity period, so
+openssl invents both: a random 159-bit serial number, and a validity
+period starting at the current time. Every signature over those fields
+differs accordingly, so two builds of the same source never produce the
+same tiboot3.bin, tispl.bin or u-boot.img on TI K3 platforms.
+
+That defeats the simplest check available to anyone re-signing a boot
+chain with their own keys: sign, then compare byte for byte against the
+original. It also leaves these images out of the reproducible builds
+support described in doc/build/reproducible.rst, even though the FIT
+wrappers around them already honour SOURCE_DATE_EPOCH.
+
+Derive the serial number from the hash of the data being signed, so
+that it is deterministic while remaining distinct for each certificate
+in an image, and keep it to 159 bits so that the encoded certificate
+does not change size. Set the validity period when SOURCE_DATE_EPOCH is
+present, running from that epoch to 20491231235959Z. Nothing which
+consumes these certificates checks the validity period: the ROM and the
+system firmware read the TI extensions carried in the certificate, and
+TI's own shipped firmware certificates have long expired.
+
+That end date is the latest which RFC 5280 section 4.1.2.5 allows to be
+encoded as UTCTime, and is deliberately not the 99991231235959Z which
+the same section offers for a certificate with no well-defined
+expiration date. The K3 boot ROM cannot parse a validity encoded as
+GeneralizedTime, which is what a date in 2050 or later requires, and
+rejects such an image outright: a Verdin AM62 (HS-FS) signed with
+99991231235959Z, or with any 2050 date, does not boot and prints
+nothing at all, while the same image signed with a 2048 or 2049 date
+boots normally. TI reports the same limitation for AM64x [1].
+
+The arguments which set the validity period require OpenSSL 3.4 or
+later, so pass them only when the available openssl is new enough and
+warn otherwise. Such a build keeps a wall-clock validity period and so
+is not reproducible, exactly as it is today. The serial number needs no
+such gate, as -set_serial has always been available.
+
+The four argument lists were byte-identical, which is how one defect
+came to be repeated four times, so factor them into a single place.
+
+[1] https://e2e.ti.com/support/processors-group/processors/f/processors-forum/1491915/am6412-ti-supports-for-tiboot3-certificates-with-validity-beyond-2049
+
+Upstream-Status: Pending
+Signed-off-by: Rogerio Guerra Borin <rogerio.borin@toradex.com>
+---
+ tools/binman/btool/openssl.py | 96 ++++++++++++++++++++++++++++++-----
+ tools/binman/ftest.py         | 32 ++++++++++++
+ 2 files changed, 116 insertions(+), 12 deletions(-)
+
+diff --git a/tools/binman/btool/openssl.py b/tools/binman/btool/openssl.py
+index c6df64c5..4ccf1fde 100644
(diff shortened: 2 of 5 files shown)


classes/include/signed/tdx-signed-k3-secboot.inc
```diff
@@ -11,6 +11,15 @@ TDX_K3_SECBOOT_KEY_DIR ?= "${TOPDIR}/keys/ti"
...
 TDX_K3_SUPPRESS_HSSE_WARNINGS ?= "0"
 
+# Whether to generate the bootloader signing certificates reproducibly, so that
+# two builds of the same sources produce byte-identical tiboot3.bin, tispl.bin
+# and u-boot.img; allowed values: 0 or 1.
+#
+# TODO: This variable is transitional. It exists only so that enabling
+#       reproducible certificates is not a breaking change within the current
+#       major release, and is to be removed in wrynose.
+TDX_K3_SECBOOT_REPRODUCIBLE ?= "0"
```
sergioprado, line 17: Is this really transitional? Do you think this would benefit all users?

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
```diff
@@ -0,0 +1,257 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
sergioprado, line 1: `From 0000000000000000000000000000000000000000` ?

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
```diff
@@ -0,0 +1,257 @@
...
++                self.assertEqual(first, second, f'{dts} is not reproducible')
++        finally:
++            if old_epoch is None:
++                del os.environ['SOURCE_DATE_EPOCH']
++            else:
++                os.environ['SOURCE_DATE_EPOCH'] = old_epoch
++
+     def testEncryptedNoAlgo(self):
+         """Test encrypted node with missing required properties"""
+         with self.assertRaises(ValueError) as e:
```
sergioprado, line 257: Can we try to upstream this patch? It does not seem to be product especifc and I guess could benefit other users.

sergioprado (changes requested):

sergioprado:
I have left a few comments. Also, please remove `Related-to: TOR-4543` from the commit message.

classes/include/signed/tdx-signed-k3-secboot.inc
rborn-tx: I expect it to benefit all users, but I wasn't sure if forcing the reproducible behavior on everyone within the same Yocto major would be acceptable. Due to this doubt, I even left the reproducibility disabled by default.

My initial use for this is to allow the validation of the new feature on TCB to re-sign bootloader binaries for the verdin-am62 (if both Yocto and TCB produce the exact same binaries we have the best possible assurance of compatibility between the two). But for that use, I'd need to set `TDX_K3_SECBOOT_REPRODUCIBLE="1"` only on test images.

So, my idea was that within `scarthgap` reproducibility would be off by default, simply because it could be considered a breaking change; but with the possibility to enable it explicitly. Then on `wrynose`, reproducibility would be always on without the possibility of disabling it (when `tdx-signed` is in use) - i.e. the variable would be dropped.

Anyway, I'm open to different views...

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
rborn-tx: Yes, my idea is to try to upstream it.

rborn-tx:
> I have left a few comments. Also, please remove `Related-to: TOR-4543` from the commit message.

Are we not using `Related-to` anymore? Or, are you asking for that because you think the commit is not totally related to that particular ticket?

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
rborn-tx: Yes, I've used `--zero-commit` when generating the patch (with `git format-patch`); the hash is not really used by OE or Git am/apply. The reason is that the commit hash would be mostly meaningless since it was created from a tree that already had some other patches applied.

Did you prefer to have a real hash here? And if so, should that hash be present upstream?

classes/include/signed/tdx-signed-k3-secboot.inc
sergioprado: Actually, my recommendation would be to remove the `TODO` and consider this the permanent behavior, with `0` as the default. Users who need it enabled (e.g. Torizon/TCB) can then enable it explicitly.

At least until this lands upstream.

What do you think?

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
sergioprado: I don't see a problem with it. We can keep it that way. It's just not very common. I searched all 5,614 patches in the layers enabled in a BSP build, and only two of them have a zeroed hash. :-)

I was just checking to make sure this was intentional rather than an error, but if it's deliberate, I'm fine with keeping it as is.

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
sergioprado: The patch looks upstreamable as it is. Why don't you submit it upstream and add an `Upstream-Status: Submitted` tag to the patch?

My intention is to follow an "upstream first" approach in this repository. I've already done the same with the patch that adapts `binman` to accept a PKCS#11 URI for HSM-backed signing.

Do you think you can do that?

sergioprado:
> Are we not using Related-to anymore? Or, are you asking for that because you think the commit is not totally related to that particular ticket?

Hmmm. This is a BSP repository, not a Torizon-specific one. I know we have used this tag in the past, but I don't think it makes much sense anymore, especially in commit messages.

Since this is an open source repository, people reading a commit with `Related-to: TOR-4543` will have no context for what that means.

And as far as I know, if the goal is simply to link the pull request to a JIRA ticket, adding `Related-to: TOR-XXXX` to the PR description should be enough.

classes/include/signed/tdx-signed-k3-secboot.inc
rborn-tx: Sounds sensible to me. I'll do that in v2.

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
rborn-tx: Ok, I'll look into it.

classes/include/signed/tdx-signed-k3-secboot.inc
rborn-tx: Done

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
rborn-tx: @sergioprado I was checking U-Boot docs and I found they usually ask for 100% coverage in binman tests.

After porting the patch to U-Boot's `master` branch and running the tests, I found that the 100% coverage level is not achieved at the moment; therefore I'd need to also augment the tests. I'd like to do that before submitting the patch upstream to avoid too many iterations.

However, I'd like to avoid keeping this PR blocked waiting for the submission. Given that, how about moving ahead with the PR? I can create a separate PR just to update the patch status once the patch is submitted.

rborn-tx:
> please remove `Related-to: TOR-4543` from the commit message.

Done

recipes-bsp/u-boot/files/0001-binman-openssl-Make-generated-certificates-reproduci.patch
sergioprado: Alright. Sounds good to me.

sergioprado (approved):
LGTM

jsrc27 (approved):

PR #192, opened 2026-09-02
Secure debug support for iMX8MP SoMs

sergioprado:
Extends the SJC backend of the secure debug feature to the iMX8M Plus, enabling it on the Verdin iMX8MP and Toradex SMARC iMX8MP.

The iMX8MP cannot offer the authenticated (Secure JTAG) mode. NXP erratum ERR052318 states that in Secure JTAG mode the SJC "does not correctly control JTAG access and may not unlock the device for JTAG access", and AN4686 Rev. 4.0 states that "the Secure Debug mode is not functional on the i.MX 8M Plus".

NXP recommends the "No Debug" mode (`TDX_SECURE_DEBUG_MODE="no-debug"`) or disabling the SJC instead (`TDX_SECURE_DEBUG_SJC_DISABLE="1"`), and that is what this PR offers.

Setting `TDX_SECURE_DEBUG_MODE = "authenticated"` on an iMX8MP fails at parsing time.

Tested on both Verdin iMX8MP and SMARC iMX8MP.

### DIFF
--- a/docs/README-secure-debug.md
+++ b/docs/README-secure-debug.md
@@ -8,7 +8,11 @@ To mitigate this risk, this layer provides a feature called **Secure Debug**.
 
 Secure Debug is currently supported on the following SoMs:
 
+- SMARC iMX8MP
 - Verdin iMX8MM
+- Verdin iMX8MP
+
+On the iMX8MP-based SoMs only the non-authenticated modes are available, because NXP defeatured the Secure JTAG mode on that SoC. See [Limitations](#limitations).
 
 Support for additional SoMs and SoC families is planned. Since each SoC family may use a different hardware mechanism to restrict debug access, new platforms may introduce additional variables or different provisioning requirements.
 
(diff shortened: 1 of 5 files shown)


docs/README-secure-debug.md
rborn-tx, line 64: Since `TDX_SECURE_DEBUG_MODE` defaults to `authenticated`, on the iMX8MP if the user only adds the inherit line to their `local.conf` they would immediately bump into the fatal error, right?

Not such a big deal, but just wanted to point it out in case you didn't realize it.

rborn-tx (commented):
Raised a minor point.

jsrc27 (approved):

docs/README-secure-debug.md
sergioprado: > Since TDX_SECURE_DEBUG_MODE defaults to authenticated, on the iMX8MP if the user only adds the inherit line to their local.conf they would immediately bump into the fatal error, right?

Yes, and that is by design. I don't want users to enable this feature without being aware that authenticated debug mode is not available on the i.MX8MP.

They can then explicitly choose between "No debug" mode or disabling SJC, but I want that choice to be deliberate rather than happening implicitly.

rborn-tx (approved):

PR #194, opened 2026-09-14
u-boot: update Upstream-Status of certificate reproducibility patch

rborn-tx:
Update Upstream-Status of certificate reproducibility patch after sending the patch upstream.


sergioprado (approved):
Nice! Since this is a pure "cosmetic" change that does not affect functionality, I will go ahead and merge it.

PR #195, opened 2026-09-15
Add Secure Debug support for iMX8/iMX8X-based SoMs

sergioprado:
This PR extends the Secure Debug feature to Apalis iMX8 and Colibri iMX8X.

These SoCs also use the System JTAG Controller (SJC), but the mechanism differs from the older iMX6, iMX7, and iMX8M families. On iMX8/iMX8X, the debug policy is tied to the AHAB lifecycle, and authentication uses two separate 128-bit keys: one for the normal world and an optional one for the secure world/TrustZone.

The user-facing interface remains mostly unchanged, with the same `authenticated` and `no-debug` policies, plus the existing full JTAG disable option. The only new variable is the optional `TDX_SECURE_DEBUG_TZ_KEY_FILE`, used to provide the TrustZone key. When it is not set, the secure-world challenge/response path is disabled, so it is never left protected by an unprogrammed key.

The fuse commands generated for iMX6, iMX7, and iMX8M are unchanged. The documentation has also been updated to describe the iMX8/iMX8X flow.

Build-tested on Apalis iMX8 and Colibri iMX8X. I also tested Colibri iMX6 and Verdin iMX8MM to check for regressions.

Runtime testing is still pending, which is why this PR is currently marked as Draft.


jsrc27 (approved):
LGTM

rborn-tx:
@sergioprado Let me ask you: what line width are you using with the commit messages body? The commit message bodies in this PR don't seem to follow the default 72 or 68 columns we usually use in Toradex's repositories. Our layer here doesn't specify the expected width, so I cannot say some convention is not being followed, but I'd expect us to follow one of the default ones.

Out of curiosity, I got these stats:

| Length | Lines |
|---:|---:|
| 68 | 162 |
| 69 | 210 |
| 70 | 250 |
| 71 | 208 |
| 72 | 208 |
| 73 | 18 |
| 74 | 9 |
| 75 | 15 |
| 76 | 9 |

From where I conclude we've been following mostly the 72-column limit...

sergioprado:
> @sergioprado Let me ask you: what line width are you using with the commit messages body? The commit message bodies in this PR don't seem to follow the default 72 or 68 columns we usually use in Toradex's repositories. Our layer here doesn't specify the expected width, so I cannot say some convention is not being followed, but I'd expect us to follow one of the default ones.

@rborn-tx It should be 72 collums. I checked the commit messages, and none exceeded that limit, but a few were wrapped earlier than necessary. I’ve rewrapped those and updated the PR. Thanks for the review!

rborn-tx (approved):
No findings on my side. Great work!

